PR #359, opened 2014-12-17
Checks for SLES12 to enable Storage repositories

dguitarbite:


toabctl:
More comments here? Would be good to get this merge to continue with the Storage product integration.

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -101,6 +127,8 @@ def get_repos(provisioner_server_node, platform, version)
             )
           when "12.0"
             repo_names = %w(
+              SUSE-Storage-1.0-Pool
+              SUSE-Storage-1.0-Updates
```
vuntz: I don't think these two lines are correct. They mean that the repos are always required, which defeats the rest of this pull request.

dguitarbite: As per my previous discussions, I thought this was the requirement. I will talk to Dirk once.

dirkmueller: No, the Storage dependency is optional, its only required for Ceph deployments.

dguitarbite:
@toabctl This code should work. If you want to test this, I suggest you to use the flag `NOQACROWBARDOWNLOAD=1` and also to edit this line https://github.com/SUSE-Cloud/automation/blob/master/scripts/qa_crowbarsetup.sh#L1227 which unsets variable `wantceph` bydefault.

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -52,24 +62,40 @@ def inspect_repos(node)
...
-            node.set[:provisioner][:suse][:missing_hae] = missing
-            node.save
+          if node[:provisioner][:suse][:missing_hae] != missing_hae
+            node.set[:provisioner][:suse][:missing_hae] = missing_hae
+          end
+          if node[:provisioner][:suse][:missing_storage] != missing_storage
+            node.set[:provisioner][:suse][:missing_storage] = missing_storage
           end
+
+          node.save
```
vuntz: Can you save only if there was a `node.set`? (Just use a variable to know if there's a need for saving)

dguitarbite: Yes, I will update the code.

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -52,22 +62,41 @@ def inspect_repos(node)
...
           # set an attribute about missing repos so that cookbooks and crowbar
           # know that HA cannot be used
+          # know that SUSE_Storage cannot be used
+          node_set=false
           node.set[:provisioner][:suse] ||= {}
-          if node[:provisioner][:suse][:missing_hae] != missing
-            node.set[:provisioner][:suse][:missing_hae] = missing
+          if node[:provisioner][:suse][:missing_hae] != missing_hae
+            node.set[:provisioner][:suse][:missing_hae] = missing_hae
+            node_set=true
```
vuntz: Style nit-picking, but this should be `node_set = true` (spaces around `=`)

dguitarbite: Thanks for catching that, my bad.

vuntz:
+1

dirkmueller:
+1

PR #362, opened 2015-01-08
Unbreak Rails 4.x port

dirkmueller:
It was not possible to join any nodes anymore..


vuntz:
So we move to ruby2.1 everywhere. Fun.

+1

PR #363, opened 2015-01-12
Do not hardcode the requirement for the cstruct rubygem anymore

vuntz:
https://github.com/crowbar/barclamp-deployer/pull/226 should obsolete
this.


dirkmueller:
+1

PR #364, opened 2015-01-12
Fix name of rest-client rubygem package

vuntz:


dirkmueller:
this is not correct.. the names differ between sle11 and sle12. perhaps we should simply fix the naming on sle11 to make it match the sle12 variant.

dirkmueller:
I fixed the sle11 naming, so this PR is no longer necessary.

PR #365, opened 2015-01-15
Add chef-client to X-Start-Before of crowbar_join init script

vuntz:
We want to start crowbar_join before chef-client.


bmwiedemann:
+1 (assuming it is tested)

vuntz:
It was tested, yes (and we even ship it) => merging.

PR #366, opened 2015-01-16
Fix target_platform attribute not being set on some nodes

vuntz:


dirkmueller:
+1

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -158,6 +158,9 @@
         os=mnode[:target_platform]
         if os.nil? or os.empty?
           os = node[:provisioner][:default_os]
+          # save the target platform we'll use
+          mnode[:target_platform] = os
+          mnode.save
```
vuntz: @rhafer mentioned that he'd prefer that in the allocate method of the crowbar app. However, in https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/models/node_object.rb#L489 we are careful to not use the node (to avoid races), so I'm not sure we really want to move this code there.

Opinions?

jsuchome:
+1

dirkmueller:
this caused https://bugzilla.suse.com/show_bug.cgi?id=913867

PR #367, opened 2015-01-20
Exclude SLE12 platform.

aplanas:
Depends on crowbar/barclamp-crowbar#1175

### DIFF
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -28,12 +28,18 @@ class ProvisionerService < ServiceObject
         "provisioner-server" => {
           "unique" => false,
           "count" => 1,
-          "admin" => true
+          "admin" => true,
+          "exclude_platform" => {
+            "suse" => "12.0"
+          }
         },
         "provisioner-base" => {
           "unique" => false,
           "count" => -1,
-          "admin" => true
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -28,12 +28,18 @@ def role_constraints
...
+            "suse" => "12.0"
+          }
         },
         "provisioner-base" => {
           "unique" => false,
           "count" => -1,
-          "admin" => true
+          "admin" => true,
+          "exclude_platform" => {
+            "suse" => "12.0"
```
vuntz: This is not correct: this is a basic role that all nodes need.

jsuchome:
No, crowbar-related barclamps are for all nodes

aplanas:
Close because is a crowbar related barclamp.

vuntz:
Well, the provisioner-server bit was relevant, imho.

aplanas:
Well, this is a bit confusing.  Maybe I can create a list of roles (barclamp-*/chef/roles/*) and decide the constraints here, so I can have a better picture.

vuntz:
@aplanas you can see all roles in /opt/dell/chef/roles on any install ;-) But yes, we should probably have a table somewhere to document this before you create all the PR.

PR #369, opened 2015-01-21
Fix target_platform attribute not being set on some nodes (second try)

vuntz:
(marked as WIP because untested)

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -158,9 +158,6 @@ if not nodes.nil? and not nodes.empty?
         os=mnode[:target_platform]
         if os.nil? or os.empty?
           os = node[:provisioner][:default_os]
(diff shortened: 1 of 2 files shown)


vuntz:
Tested

jsuchome:
It might be really needed for "hardware-installing" only, but wouldn't it be better to move it up,before even checking the state?

vuntz:
@jsuchome I think the goal was to not do it unless a node is allocated, and hardware-installing is the first state that we reach after allocating a node. I guess an alternative is to check if the node is allocated and then do this.

vuntz:
Clicked "comment" to fast: would you prefer me to use the alternative I mentioned above?

jsuchome:
I do not really have strong opinion here, that was just something that came to mind first. I think the patch looks good (but I thought the same about previous one)

vuntz:
Merging as there was no objection in two weeks, it fixes a P1 bug and this has been tested as package patch.

PR #375, opened 2015-01-30
Keep a note about OS which we know how to install but are not available

vuntz:
Right now, as soon as we know about an OS, we pretend we can install it.
But if it's missing some files that the sysadmin needs to put on the
admin server, the OS is not yet 100% available.

So we keep a note that this OS should be disabled (so that we can show
it can be made available, although not totally available right now).


tboerger:
LGTM

PR #377, opened 2015-02-03
Do not rely on DMI information being available

dirkmueller:
On newer hardware or when installing the admin node in the Xen PVM
there is no DMI information available, so do not rely on it.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/bootdisk.rb
+++ b/chef/cookbooks/provisioner/recipes/bootdisk.rb
@@ -28,8 +28,8 @@ end
 # Unless some other barclamp has set it, that is.
 ruby_block "Find the fallback boot device" do
   block do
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/recipes/bootdisk.rb
```diff
@@ -57,7 +57,8 @@
 
     basedir="/dev/disk/by-id"
     # /dev/disk/by-id is unstable under VirtualBox so don't rely on it.
-    if node[:dmi][:system][:product_name] !~ /VirtualBox/i && File.exists?(basedir)
+    hardware=node[:dmi][:system][:product_name] rescue "unknown"
```
tboerger: Spaces before and after ```=```?

vuntz:
+1 once style nitpick is fixed

PR #378, opened 2015-02-05
Stop crowbar_join before openvswitch and ovs-usurp-config-br-fixed

vuntz:
When stopping after these services, crowbar_join is not able to reach
the admin server to change the state of the node.


rhafer:
Makes sense. Though shouldn't we also add them to ```Should-Start```. Otherwise there might be a race I think. E.g. crowbar_join starting in between openvswitch and the usurp script.

dirkmueller:
+1

PR #379, opened 2015-02-05
Fix crowbar_register on SLE12

vuntz:
Do not install suse-cloud-release but suse-sle12-cloud-compute-release.

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -177,12 +177,19 @@ PATTERNS_INSTALL="$PATTERNS_INSTALL Minimal base"
 PACKAGES_INSTALL="$PACKAGES_INSTALL biosdevname netcat ruby2.1-rubygem-chef supportutils-plugin-susecloud"
 
 # From sleshammer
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -182,7 +182,11 @@ PACKAGES_INSTALL="$PACKAGES_INSTALL parted lvm2 ipmitool curl tack nfs-client rs
 
 zypper --non-interactive install -t pattern $PATTERNS_INSTALL
 # Auto-agree with the license since it was already agreed on for the admin server
+<% if @os == "suse-11.3" -%>
```
dirkmueller: previously we used

<% if @target_platform_version == "11.3" -%>

MaximilianMeister:
tested (although not within barclamp context, just `crowbar_register`)
+1

tboerger:
Why do you install som gems only on sle < 12?

dirkmueller:
+1

vuntz:
> Why do you install som gems only on sle < 12?

I'm actually thinking of just removing that line as it's only needed when doing a hardware update, I think. But I went the safe way and left that for SLE11.

tboerger:
So these gems are only used on older SLES versions?

dirkmueller:
Improves things, so merging.

vuntz:
@tboerger as I said, I think they're actually not used at all. But I don't feel comfortable merging this now. Let me add a new PR for post-tex.

tboerger:
Too late, @dirkmueller already merged that

PR #381, opened 2015-02-09
[post-tex] crowbar_register: Do not install packages that are in sleshammer

vuntz:
If they are needed by the cookbooks, then the cookbooks should install
the package.


tboerger:
But crowbar_register gets called for non sleshammer instances as well or not?

vuntz:
@tboerger it's never called from sleshammer, always from non-sleshammer. But the difference compared to sleshammer is that cookbooks can install packages now. And they should do that instead of relying on the packages to be installed.

tboerger:
:+1:

rhafer:
@vuntz: Please rebase

vuntz:
Rebased

rsalevsky:
:+1:

PR #383, opened 2015-02-16
only try to disable SuSEfirewall if installed

bmwiedemann:


jdsn:
:+1:  removes dependency to SuSEfirewall

PR #384, opened 2015-02-16
crowbar_register fixes

vuntz:
- Fail early if we can't communicate with admin server
- Do not do state transitions if node is registered

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -165,49 +165,42 @@ fi
 # Helper functions
 # ----------------
 
+get_common_curl_args() {
+  local curlargs=(--connect-timeout 60 -s -L --max-time 240)
+
+  [[ $CROWBAR_KEY ]] && curlargs+=(-u "$CROWBAR_KEY" --digest --anyauth)
+
+  echo ${curlargs[@]}
+}
+
 get_nodes_allow_fail() {
   # --fail is important here as we rely on the failure to know if we can
   # communicate with the admin server
-  local curlargs=(--connect-timeout 60 -s --fail \
-                  -L
-                  -H "Accept: application/json" \
-                  --max-time 240)
-
-  [[ $CROWBAR_KEY ]] && curlargs+=(-u "$CROWBAR_KEY" --digest --anyauth)
+  local curlargs=($(get_common_curl_args) --fail -H "Accept: application/json")
 
   curl "${curlargs[@]}" "http://$ADMIN_IP:3000/crowbar/machines/1.0/"
 }
 
 get_node_allow_fail() {
   # --fail is important here as we rely on the failure for 404
-  local curlargs=(--connect-timeout 60 -s --fail \
-                  -L
-                  -H "Accept: application/json" \
-                  --max-time 240)
-
-  [[ $CROWBAR_KEY ]] && curlargs+=(-u "$CROWBAR_KEY" --digest --anyauth)
+  local curlargs=($(get_common_curl_args) --fail -H "Accept: application/json")
 
   curl "${curlargs[@]}" "http://$ADMIN_IP:3000/crowbar/machines/1.0/$HOSTNAME"
 }
 
 post_allocate() {
-  local curlargs=(--connect-timeout 60 -s \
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -283,6 +296,18 @@ curl -L -o /etc/chef/validation.pem \
...
     "$HTTP_SERVER/validation.pem"
 
+nodes=$(get_nodes_allow_fail || echo "")
+if test -z "$nodes"; then
+  echo "Failed to list nodes known on the administration server."
+  echo "Is the administration server up?"
+  exit 1
+fi
+
+if echo "$nodes" | grep -q "\"name\":\"$HOSTNAME\""; then
```
dirkmueller: it feels simpler to me to just query for 

/crowbar/machines/1.0/$HOSTNAME

is also slightly faster (show instead of list I guess)

vuntz: Well, I still need an URL that always works for the first test. I could send another request for this test, but then it's another request ;-)

tboerger: And anyway it's bad usage of grep or not? Maybe ```grep -q "\"name\":\"$HOSTNAME\"" "$nodes"```?

vuntz: @tboerger: err, no. `"$nodes"` is not a file; this needs to be written to stdin to have grep work.

tboerger: Ok, than ```grep -q "\"name\":\"$HOSTNAME\"" <<< $nodes``` :)

aspiers: Yuk!

    if [[ "$nodes" == *\"name\":\""$HOSTNAME"\"* ]]; then

aspiers: Actually I don't think you even need to quote the variables:

    if [[ $nodes == *\"name\":\"$HOSTNAME\"* ]]; then

dirkmueller: @vuntz the url I proposed always works.. it will either return a json blob or a (html) error page. if it doesn't return anything the chef server is unreachable..

vuntz: There you go. Should be better.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -147,6 +147,77 @@ if test "x$MD5_ADMIN" != "x$MD5_LOCAL"; then
...
+}
+
+get_node_allow_fail() {
+  # --fail is important here as we rely on the failure for 404
+  local curlargs=(--connect-timeout 60 -s --fail \
+                  -L
+                  -H "Accept: application/json" \
+                  --max-time 240)
+
+  [[ $CROWBAR_KEY ]] && curlargs+=(-u "$CROWBAR_KEY" --digest --anyauth)
```
toabctl: The ```curlargs``` preparation is redundant in ```get_nodes_allow_fail``` and ```get_node_allow_fail```. Maybe add that to an extra function?

rhafer:
@vuntz: Please rebase

vuntz:
Rebased

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
rhafer: I guess both functions could also be merged, by just allowing to pass ```hostname``` as a parameter, no?

rsalevsky:
@vuntz: Rebase please

vuntz:
Rebased

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
vuntz: Done, but not tested yet.

rsalevsky: `get_common_curl_args()` returns only `--connect-timeout`

rsalevsky:
+1 tested and works for me

PR #386, opened 2015-02-18
sledgehammer: Tiny fix to output

vuntz:
We were missing a newline character between "echo -n" and a new line
output by another echo.


aspiers:
+1

PR #387, opened 2015-02-18
crowbar_register: Set the target platform of the node we're registering

vuntz:
Otherwise, it'll default to the default of the admin server, which might
not be correct (ie, not SLE12).

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -285,8 +285,13 @@ curl -L -o /etc/chef/validation.pem \
     --connect-timeout 60 -s \
     "$HTTP_SERVER/validation.pem"
 
+TMP_ATTRIBUTES=$(mktemp --suffix .json)
+
+# Make sure that we have the right target platform
+echo "{ \"target_platform\": \"$CROWBAR_OS\" }" > "$TMP_ATTRIBUTES"
+
 post_state $HOSTNAME "discovering"
-chef-client -S http://$ADMIN_IP:4000/ -N "$HOSTNAME"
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
so this sets the target platform in the UI -> node view?

vuntz:
@MaximilianMeister this sets the attribute that the UI is using. But other bits are using this attribute (some cookbooks for instance). That's why it's important to get it right.

MaximilianMeister:
good to know.
+1

tboerger:
LGTM

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -285,8 +285,13 @@ curl -L -o /etc/chef/validation.pem \
     --connect-timeout 60 -s \
     "$HTTP_SERVER/validation.pem"
 
+TMP_ATTRIBUTES=$(mktemp --suffix .json)
+
+# Make sure that we have the right target platform
+echo "{ \"target_platform\": \"$CROWBAR_OS\" }" > "$TMP_ATTRIBUTES"
```
dirkmueller: does that add it as normal attributes or as default attribute ? in mkcloud we're setting normal.target_platform via knife node edit

vuntz: Hrm, I don't know, but I would expect normal attribute. Would need to try again to be sure.

MaximilianMeister: > does that add it as normal attributes or as default attribute ? in mkcloud we're setting normal.target_platform via knife node edit

its added as "normal", i just checked

PR #388, opened 2015-02-18
crowbar_register: Do not run if connected through ssh with no screen

vuntz:
There'll be a network reconfiguration that will kill the script...

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -63,6 +63,12 @@ if [ $(id -u) -gt 0 ]; then
     exit 1
 fi
 
+if test -n "$SSH_CONNECTION" -a -z "$STY"; then
+    echo "Not running in screen. Please run $0 inside screen to avoid problems during network re-configuration."
+    echo ""
+    exit 1
+fi
+
 if test $CROWBAR_FORCE -ne 1; then
     echo    "Running this tool will alter the system for integration with SUSE Cloud."
     echo -n "Continue? [y/N]: "


aspiers:
+1.  Nice idea to check `$SSH_CONNECTION`; should we also do that in `install-chef-suse.sh` ?

MaximilianMeister:
good idea indeed. especially because this is not mentioned in the documentation
+1

vuntz:
@aspiers yes, I think the check in install-chef-suse.sh should be enhanced like this. Want to do it? Or shall I?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -63,6 +63,12 @@ if [ $(id -u) -gt 0 ]; then
     exit 1
 fi
 
+if test -n "$SSH_CONNECTION" -a -z "$STY"; then
```
aspiers: Hmm, it seems there are roughly an equal number of usages of `if test -...` and `if [ ]` / `if [[ ]]`.  I wish we could be consistent about this :-/  My take is that `test` is the more old-fashioned style, from the POSIX era where it actually forked a separate process.  Personally I like the conditional to be wrapped in parentheses, especially when it spans multiple lines.

aspiers:
@vuntz Done: https://github.com/crowbar/crowbar/pull/2089

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
dirkmueller: test and [ is identical with any current shell performance wise. as you noticed, there is no predominant style yet so it doesn't make sense to refuse the PR on that base yet.

dirkmueller:
+1

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
aspiers: Yes, performance and behaviour are identical, I was talking purely about style.  However even though there was no predominant style within this particular file, the predominant style in the rest of SUSE Cloud is to use `[ ]`.  So we should try be consistent across the team and product.  Although on this occasion I didn't write "-1".

PR #391, opened 2015-02-20
Allow another architectures to be registered

k0da:
Currently crowbar_register assumes we're running on x86_64 only.
If we want to add other KVM capable architectures, it is simply not
possible.

Let's change it in a way that we separate architectures specific
packages in to arch dependent list and add arch suffix into a repo name.

Signed-off-by: Dinar Valeev <dvaleev@suse.com>

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -153,6 +153,10 @@ if test "x$MD5_ADMIN" != "x$MD5_LOCAL"; then
 fi
 
 
+#we need to know an architecture we are running on
+ARCH=`uname -m`
+
+
 # Setup the repos
 # ---------------
 
@@ -184,7 +188,7 @@ PACKAGES_INSTALL="$PACKAGES_INSTALL openssh"
 
 # From autoyast profile
 PATTERNS_INSTALL="$PATTERNS_INSTALL Minimal base"
-PACKAGES_INSTALL="$PACKAGES_INSTALL biosdevname netcat ruby2.1-rubygem-chef supportutils-plugin-susecloud"
+PACKAGES_INSTALL="$PACKAGES_INSTALL netcat ruby2.1-rubygem-chef supportutils-plugin-susecloud"
 
 # From sleshammer
 PACKAGES_INSTALL="$PACKAGES_INSTALL ruby2.1-rubygem-chef ruby2.1-rubygem-rest-client ruby2.1-rubygem-cstruct"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -152,19 +152,21 @@ if test "x$MD5_ADMIN" != "x$MD5_LOCAL"; then
     exit 1
 fi
 
+#we need to know an architecture we are running on
+ARCH=`uname -m`
```
vuntz: You need to add an empty line after this one, to keep the two empty lines before the next code section.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -152,19 +152,21 @@ if test "x$MD5_ADMIN" != "x$MD5_LOCAL"; then
...
+zypper -n ar "<%= @repos[name][:url] %>" "<%= name %>-$ARCH"
 <% end %>
 
 # Cloud-PTF has an unknown key, and this is expected
 <% if @target_platform_version.to_f < 12.0 -%>
-zypper -n --gpg-auto-import-keys refresh -r SLE-Cloud-PTF
+zypper -n --gpg-auto-import-keys refresh -r SLE-Cloud-PTF-$ARCH
 <% else -%>
-zypper -n --gpg-auto-import-keys refresh -r SLE12-Cloud-Compute-PTF
+zypper -n --gpg-auto-import-keys refresh -r SLE12-Cloud-Compute-PTF-$ARCH
```
vuntz: So this bit, while it makes sense, probably needs some deep care.

I think we have code that periodically think the repos configured on the nodes, and this definitely doesn't know about the architecture.

Can you possibly make this pull request only about the biosdevname bit, which we could easily merge, and then create another pull request for the repo change?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -193,6 +195,10 @@ PACKAGES_INSTALL="$PACKAGES_INSTALL ruby2.1-rubygem-libxml-ruby ruby2.1-rubygem-
 <% end -%>
 PACKAGES_INSTALL="$PACKAGES_INSTALL parted lvm2 ipmitool curl tack nfs-client rsyslog ntp tcpdump"
 
+case $ARCH in
+    x86_64) PACKAGES_INSTALL+=" biosdevname";;
```
vuntz: Is the package only available in x86_64, or is it not available on your arch but available on the other ones? Just wondering if this should be something like:
```shell
case $ARCH in
    ppc*) ;;
    *) PACKAGES_INSTALL+=" biosdevname";;
esac

k0da: bios says this is a package to work on BIOS systems which are x86

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
k0da: Yes, qa_crowbarsetup needs to adapt to that change. I could easily make change to biosdevname only.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
k0da: Ok will do it

tboerger:
:+1:

vuntz:
+1

dirkmueller:
+1

PR #392, opened 2015-02-23
Configuration error. Should be 'root_password_hast', see bc-template-pro...

rrmichel:
Here is a mismatch between the variable defined in the schema (root_password_hash) and in the autoyast file (rootpw)

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -44,8 +44,8 @@
   <users config:type="list">
       <user>
          <username>root</username>
-<% if @rootpw_hash != "" %>
-         <user_password><%= @rootpw_hash %></user_password>
+<% if @root_password_hash != "" %>
+         <user_password><%= @root_password_hash %></user_password>
          <encrypted config:type="boolean">true</encrypted>
 <% end %>
          <forename/>


dirkmueller:
@rrmichel you also need to update chef/cookbooks/provisioner/recipes/update_nodes.rb then to set root_password_hash instead of rootpw_hash. 

In general I think you're trying to fix a bug but this is not it... can you elaborate what the issue is?

rrmichel:
Try to set the 'root_default_hash' over crowbar, install a machine (with the autoyast file) and try to login.

After the installation i found a blank password for root

<username>root</username>
<user_password></user_password>

Maybe your right and i missed some code....

vuntz:
@rrmichel how di you set "the 'root_default_hash' over crowbar"? Did you apply the provisioner proposal?

rrmichel:
@vuntz not 'root_password_hash' ?? I've set 'root_password_hash' and applyed the proposal.

vuntz:
@rrmichel sorry, I pasted the attribute name from your comment ;-)

So if you look at the code at https://github.com/crowbar/barclamp-provisioner/blob/master/chef/cookbooks/provisioner/recipes/update_nodes.rb#L243 you can see that the attribute is being used, and is used to set the `@rootpw_hash` variable in the template. That's the part I don't understand about your patch.

rrmichel:
@vuntz i already take notice about the code....but if set the 'rootpw_hash' in den provisioner proposal, i get a autoyast configuration with an emtpy password for root.

But if i patch the autoyast erb file, set the root_password_hash in the proposal, it works!

I will check this again...

rrmichel:
Gnarf....was an error on my configuration...copy'n'paste error - sorry for

PR #393, opened 2015-02-23
Don't install ruby 1.8 anymore on the nodes

dirkmueller:
We don't need it anymore.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -252,6 +252,13 @@ sync
 <% end -%>
       <package>supportutils-plugin-susecloud</package>
     </packages>
+<% if @target_platform_version == "11.3" -%>
+    <remove-packages config:type="list">
+      <package>puppet</package>
+      <package>facter</package>
+      <package>ruby</package>
+    </remove-packages>
+<% end -%>
     <patterns config:type="list">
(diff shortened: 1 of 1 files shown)


dirkmueller:
it has the slight downside that the installer comes up in text mode now.. not sure how to resolve that.

tboerger:
LGTM

vuntz:
I'm wondering if it makes a big difference?

dirkmueller:
@vuntz it avoids that something accidentally refers to /usr/bin/ruby and we git strange bugs due to that. it also removes a few MB of disk usage, but I guess you're not worried about that ;)

tboerger:
We have puppet and facter installed somewhere?

toabctl:
+1 (but not tested)

dirkmueller:
@tboerger puppet is installed on all machines by the base pattern. thats why I need to add the override in order to be able to not install ruby 1.8

tboerger:
Ok :+1:

rsalevsky:
+1

aspiers:
It would be nice if the commit message briefly explained why it is no longer needed.

rsalevsky:
@dirkmueller, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).
Please open a new pull request there. __Do NOT merge here anymore!__

dirkmueller:
don't care about it anymore

PR #397, opened 2015-02-25
Display an error when DHCPServer is not found  (bnc#914063)

ovisan:
Added Messagebox functionality to display error messages as popups. The
code will not continue until the OK button on the popup is pressed.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/set_state.ps1.erb
+++ b/chef/cookbooks/provisioner/templates/default/set_state.ps1.erb
@@ -1,3 +1,6 @@
+#Loading windows forms for MessageBox popup
+[System.Reflection.Assembly]::LoadWithPartialName("System.Windows.Forms")
+
 Param(
   [string]$CrowbarAdminIP = "<%= @admin_ip %>",
   [string]$CrowbarKey = "<%= @crowbar_key %>"
@@ -13,6 +16,9 @@ foreach ($card in $cards)
     $hostname = "d$($hostname.ToLower())"
     break
   }
+}
+if (!$hostname) {
+   [System.Windows.Forms.MessageBox]::Show("DHCPServer Not Found $CrowbarAdminIP", "Error", 0, 16)
 }
 
 $uri =  "http://"+$CrowbarAdminIP+":3000"


bmwiedemann:
:+1:  - note that code can not continue without the missing value, so it is the best to stop there with a useful message for the user

vuntz:
I'm not sure I understand: when can this happen? Surely, if we reached this point, then we had a DHCP server at some point?

ovisan:
https://trello.com/c/EN2IqC2W/247-5-p7-bnc-914063-hyper-v-server-2012-r2-installation-loop-error
added for tracking

vuntz:
> I'm not sure I understand: when can this happen? Surely, if we reached this point, then we had a DHCP server at some point?

@ovisan @bmwiedemann any answer to this?

ovisan:
By DHCP server it means the crowbar node. If the windows machine cannot find it (via the powershell script) it will reboot, hence the reboot loop bug....
This "fix" is intended to display an error message in windows so that it will be visible via ILO, BMC... and the node will not reboot until the user will press OK.

vuntz:
But my question is again: why can't it find it? It seems to me we're fixing a symptom, not the bug.

tboerger:
Codewise it LGTM, but still the question of @vuntz :)

rsalevsky:
@ovisan Is this PR still required? When yes please answer the open question.

ovisan:
@rsalevsky @tboerger @vuntz , the "DHCP server"  in this case is the crowbar node. And to be more explicit, this happens IF there are 2 ip's in the installation process, the powershell script doesn't know which one to choose. Of course i could have modified the powershell script, but as i understood, having 2 ip's it's wrong anyway, so this change just display's an error message on the windows side stating that there is no "DHCP server" found instead of looping forever and having the client clueless.
So, in short, yes, this PR is needed.

rsalevsky:
Thanks, for me it makes sense now. :+1:

vuntz:
> Of course i could have modified the powershell script, but as i understood, having 2 ip's it's wrong anyway (but hey, it can happen, right?)

That's really the part I'm not sure I understand and that is blocking me. What does "the admin server has 2 IP addresses" mean? Two addresses on one interface? Two addresses in general?

Also, if it can happen, then we should just support it and not error out, especially if the error doesn't really help people know how to fix things.

ovisan:
@vuntz for this 2 IP's issue, @bmwiedemann can provide more details

mjura:
I experienced problem described by @ovisan, it would be useful to have this change. It is needed to debug Windows integration.

For me +1

rsalevsky:
@ovisan, this repository got merged into [crowbar-core](https://github.com/crowbar/crowbar-core).
Please open a new pull request there. __Do NOT merge here anymore!__

PR #398, opened 2015-02-25
Implement keep_existing_hostname option

rsalevsky:
This option implements possibility to keep the existing hostname.

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -22,7 +22,7 @@
 
 usage () {
     cat <<EOF
-`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [--no-gpg-checks] [--interface IF]
+`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [-k|--keep-existing-hostname] [--no-gpg-checks] [--interface IF]
 
 Register node in Crowbar.
 EOF
@@ -34,6 +34,11 @@ CROWBAR_AUTO_IMPORT_KEYS=0
 CROWBAR_NO_GPG_CHECKS=0
 CROWBAR_FORCE=0
 CROWBAR_VERBOSE=0
+<% if !node[:provisioner][:keep_existing_hostname] -%>
+KEEP_EXISTING_HOSTNAME=0
+<% else %>
+KEEP_EXISTING_HOSTNAME=1
+<% end %>
 DEFINEDDEV=
 
 while test $# -gt 0; do
@@ -41,6 +46,7 @@ while test $# -gt 0; do
         -h|--help|--usage|-\?) usage ;;
         -v|--verbose) CROWBAR_VERBOSE=1 ;;
         -f|--force) CROWBAR_FORCE=1 ;;
+        -k|--keep-existing-hostname) KEEP_EXISTING_HOSTNAME=1 ;;
         --gpg-auto-import-keys) CROWBAR_AUTO_IMPORT_KEYS=1 ;;
         --no-gpg-checks) CROWBAR_NO_GPG_CHECKS=1 ;;
         --interface)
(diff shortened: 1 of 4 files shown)


vuntz:
Hrm, any reason to use an attribute for that? I would expect this to be a command line option for crowbar_register instead.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -260,7 +260,15 @@ post_state() {
...
+if ! echo "$HOSTNAME" | grep -q -o "$DOMAIN$"; then
+  echo "The fully qualified domain name did not contain the cloud domain."
+  exit 1
+fi
+<% else -%>
 HOSTNAME="d${MAC//:/-}.${DOMAIN}"
+<% end -%>
 sed -i -e "s/\(127\.0\.0\.1.*\)/127.0.0.1 $HOSTNAME ${HOSTNAME%%.*} localhost.localdomain localhost/" /etc/hosts
 echo "$HOSTNAME" > /etc/HOSTNAME
 hostname "$HOSTNAME"
```
vuntz: It's unclear to me if we want to do these last 3 lines when we preserve the hostname. I would expect that we wouldn't want that, but I'm not 100% sure.

There's also https://github.com/rsalevsky/barclamp-provisioner/blob/hostname/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb#L329 to consider

rsalevsky:
I discussed this with Dirk, the reason for this decision is that the customer want's to keep all hostnames. This way avoids the fault of forgetting the option and it is more comfortable.

vuntz:
@rsalevsky I must admit I find this wrong somehow. Would be interesting to hear from others.

An alternative is to make this an option in the script, but also have an attribute control what the default is for the option. Might be too much, though.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
rsalevsky: Normally we can skip this lines but in this case they didn't cause any unwanted changes.

rsalevsky:
Yes, this can be the best solution between comfort and flexibility. But let's wait for other feedback/ideas.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
vuntz: Well, it's wrong to rewrite `/etc/HOSTNAME` for no reason :-)

toabctl:
I may misunderstand the function of ```crowbar_register```, but isn't the intention of that script to register a new node to crowbar? Given that I have a fresh node which has nothing todo with Crowbar for now (but it's in the same subnet so it can reach the admin server) - how do I use that script from the node to add it to crowbar? Currently it's a template but shouldn't it be a file just accepting some parameters?

rsalevsky:
The script template is rendered on every barclamp-provisioner apply run and is stored below `/srv/tftpboot/suse-12.0/crowbar_register` or `/srv/tftpboot/suse-11.3/crowbar_register`. You can download it from the crowbar server via port 8091.

toabctl:
> The script template is rendered on every barclamp-provisioner apply run and is stored below /srv/tftpboot/suse-12.0/crowbar_register or /srv/tftpboot/suse-11.3/crowbar_register. You can download it from the crowbar server via port 8091.

Ah. Thanks for the explanation. Then I like @vuntz idea to have both - a global switch to keep all hostnames via an attribute but also a CLI parameter when the global switch is not used.

rhafer:
+1

tboerger:
Do we want to expose that in the UI? Or do we want to keep it in the raw view only?
:+1:

rhafer:
> Do we want to expose that in the UI? Or do we want to keep it in the raw view only?

raw view is fine for now.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -22,7 +22,7 @@
 
 usage () {
     cat <<EOF
-`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [--no-gpg-checks] [--interface IF]
+`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [-k|--keep-existing-hostname] [--no-gpg-checks] [--interface IF]
```
vuntz: I'm coming after the merge, but still some comments: maybe we should drop the option when `node[:provisioner][:keep_existing_hostname]` is set? (It doesn't make a lot of sense in that case). Or maybe this option should accept a 0/1 value?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -260,9 +266,17 @@ post_state() {
 # -----------------------
 
 MAC=$(cat /sys/class/net/$BOOTDEV/address)
-HOSTNAME="d${MAC//:/-}.${DOMAIN}"
+if test $KEEP_EXISTING_HOSTNAME -ne 1; then
+  HOSTNAME="d${MAC//:/-}.${DOMAIN}"
+  echo "$HOSTNAME" > /etc/HOSTNAME
+else
+  HOSTNAME=$(cat /etc/HOSTNAME)
+  if ! echo "$HOSTNAME" | grep -q -o "$DOMAIN$"; then
```
vuntz: Strictly speaking, you actually need to also check that there's no dot once you remove the `.$DOMAIN` suffix from `$HOSTNAME`.

(I'd also change the grep to use `.$DOMAIN$`)

PR #400, opened 2015-02-27
Add kvm and qemu user/group with fixed uid/gid

rsalevsky:
The UIDs and GIDs are diffrent between SLES12 and SLES11 SP3 to prevent
premission errors for shared folders we must create them with fixed
values.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -51,7 +51,30 @@
          <forename/>
          <surname/>
       </user>
+      <user>
+        <username>qemu</username>
+        <user_password>!</user_password>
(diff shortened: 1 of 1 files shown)


rhafer:
+1

aspiers:
+1

tboerger:
What happens with the uids and gids for nodes that are registered and not installed through autoyast?

aspiers:
Good question.  @rhafer / @rsalevsky ?

rhafer:
Hm, good point. I guess we need to handle that in crowbar_register. It will be problematic however to handle the case when the qemu user/group already exist when crowbar_register is running. I don't think there is much we can do in that case other than erroring out (and having it documented).

rhafer:
FYI: I've update the related Trello Card with additional tasks.

rsalevsky:
See #402

PR #402, opened 2015-03-04
Add kvm and qemu user/group with fixed id's

rsalevsky:
Add kvm and qemu user/group with fixed uid/gid while
crowbar_register.

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -85,6 +85,71 @@ if test $CROWBAR_FORCE -ne 1; then
 fi
 
 
+# Helper functions
+# ----------------
+
+get_common_curl_args() {
+  local curlargs=(--connect-timeout 60 -s -L --max-time 240)
+
+  [[ $CROWBAR_KEY ]] && curlargs+=(-u "$CROWBAR_KEY" --digest --anyauth)
+
+  echo ${curlargs[@]}
+}
+
+get_nodes_allow_fail() {
+  # --fail is important here as we rely on the failure to know if we can
+  # communicate with the admin server
+  local curlargs=($(get_common_curl_args) --fail -H "Accept: application/json")
+
+  curl "${curlargs[@]}" "http://$ADMIN_IP:3000/crowbar/machines/1.0/"
+}
+
+get_node_allow_fail() {
+  # --fail is important here as we rely on the failure for 404
+  local curlargs=($(get_common_curl_args) --fail -H "Accept: application/json")
+
+  curl "${curlargs[@]}" "http://$ADMIN_IP:3000/crowbar/machines/1.0/$HOSTNAME"
+}
+
+post_allocate() {
+  local curlargs=($(get_common_curl_args) \
+                  -X POST --data-binary "{ \"name\": \"$1\" }" \
+                  -H "Accept: application/json" -H "Content-Type: application/json")
+
+  curl "${curlargs[@]}" "http://$ADMIN_IP:3000/crowbar/machines/1.0/allocate/0"
+}
+
+post_state() {
+  local curlargs=($(get_common_curl_args) \
+                  -o "/var/log/crowbar/$1-$2.json"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -146,6 +146,29 @@ if test "x$MD5_ADMIN" != "x$MD5_LOCAL"; then
...
 
+# Setup groups and users
+# # ---------------
+
+if [[ -z `grep '^qemu' /etc/group` ]]; then
+  groupadd --system --gid 484 qemu
+elif [[ -z `grep '^qemu:*.:484:*.' /etc/group` ]]; then
+  echo "Group qemu did not have gid 484."
+  exit 1
+fi
```
rhafer: Please use ```getent group qemu``` that's better than grepping through /etc/group.

And I think something like this would be a bit more readable:

```
gid=$(getent group qemu | cut -f 3 -d ":")
if [ -z "$gid" ]; then
   ...create group...
elif [ "$gid" != "484" ]; then
    ...error message...
    exit 1
fi
```

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -146,6 +146,29 @@ if test "x$MD5_ADMIN" != "x$MD5_LOCAL"; then
...
+
+if [[ -z `grep '^kvm' /etc/group` ]]; then
+  groupadd --system --gid 485 kvm
+elif [[ -z `grep '^kvm:*.:485:*.' /etc/group` ]]; then
+  echo "Group kvm did not have gid 485."
+  exit 1
+fi
+
```
rhafer: Same as: https://github.com/crowbar/barclamp-provisioner/pull/402/files#r25791094

rhafer: Actually, repeating the same code three times, screems for moving it to a function :)

rhafer:
+1

toabctl:
+1

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -72,6 +72,51 @@ if test $CROWBAR_FORCE -ne 1; then
...
+  curl "${curlargs[@]}" "http://$ADMIN_IP:3000/crowbar/crowbar/1.0/transition/default"
+}
+
+add_group() {
+  gid=$(getent group $1 | cut -f 3 -d ":")
+  if [[ -z "$gid" ]]; then
+    groupadd --system --gid $2 $1
+  elif [[ "$gid" != "$2" ]]; then
+    echo "Group $1 did not have gid $2."
```
vuntz: s/did/does/

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -72,6 +72,51 @@ if test $CROWBAR_FORCE -ne 1; then
...
+    exit 1
```
vuntz: Should we exit? Or should we propose to fix it by changing the gid?

rsalevsky: I would say exit, changing it automatically can break other parts of the system due to we do not change the ID for existing files.

rhafer: Yeah, I agree. Changing the gid would require to scan the filesystem for files/dirs still owned by the old gid. Let's not go down that route.

rsalevsky:
@vuntz You have the last word.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
toabctl: I also agree and prefer to exit.

toabctl:
Still +1 . Anything left todo before we can merge this?

rsalevsky:
From my side is all done.

PR #403, opened 2015-03-05
Improve keep_existing_hostname handling in crowbar_register

rsalevsky:
Improvements for #398

### DIFF
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -22,7 +22,11 @@
 
 usage () {
     cat <<EOF
-`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [-k|--keep-existing-hostname] [--no-gpg-checks] [--interface IF]
+<% if !node[:provisioner][:keep_existing_hostname] -%>
+`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [--keep-existing-hostname] [--no-gpg-checks] [--interface IF]
+<% else -%>
+`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [--no-gpg-checks] [--interface IF]
+<% end -%>
 
 Register node in Crowbar.
 EOF
@@ -36,9 +40,9 @@ CROWBAR_FORCE=0
 CROWBAR_VERBOSE=0
 <% if !node[:provisioner][:keep_existing_hostname] -%>
 KEEP_EXISTING_HOSTNAME=0
-<% else %>
+<% else -%>
 KEEP_EXISTING_HOSTNAME=1
-<% end %>
+<% end -%>
 DEFINEDDEV=
 
 while test $# -gt 0; do
@@ -46,7 +50,9 @@ while test $# -gt 0; do
         -h|--help|--usage|-\?) usage ;;
         -v|--verbose) CROWBAR_VERBOSE=1 ;;
         -f|--force) CROWBAR_FORCE=1 ;;
-        -k|--keep-existing-hostname) KEEP_EXISTING_HOSTNAME=1 ;;
+        <% if !node[:provisioner][:keep_existing_hostname] -%>
+        --keep-existing-hostname) KEEP_EXISTING_HOSTNAME=1 ;;
+        <% end -%>
         --gpg-auto-import-keys) CROWBAR_AUTO_IMPORT_KEYS=1 ;;
         --no-gpg-checks) CROWBAR_NO_GPG_CHECKS=1 ;;
         --interface)
@@ -251,8 +257,11 @@ if test $KEEP_EXISTING_HOSTNAME -ne 1; then
   HOSTNAME="d${MAC//:/-}.${DOMAIN}"
 else
   HOSTNAME=$(cat /etc/HOSTNAME)
-  if ! echo "$HOSTNAME" | grep -q -o "$DOMAIN$"; then
-    echo "The fully qualified domain name did not contain the cloud domain."
+  if ! echo "$HOSTNAME" | grep -q "\.$DOMAIN$"; then
+    echo "The fully qualified domain name did not contain the $DOMAIN cloud domain."
+    exit 1
+  elif echo "${HOSTNAME%%.$DOMAIN}" | grep -q "\."; then
+    echo "The hostname is not a fully qualified domain name."
     exit 1
   fi
 fi


chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -46,9 +46,16 @@ while test $# -gt 0; do
         -h|--help|--usage|-\?) usage ;;
         -v|--verbose) CROWBAR_VERBOSE=1 ;;
         -f|--force) CROWBAR_FORCE=1 ;;
-        -k|--keep-existing-hostname) KEEP_EXISTING_HOSTNAME=1 ;;
         --gpg-auto-import-keys) CROWBAR_AUTO_IMPORT_KEYS=1 ;;
         --no-gpg-checks) CROWBAR_NO_GPG_CHECKS=1 ;;
+        -k|--keep-existing-hostname)
+            if test $# -eq 1; then
```
vuntz: Unfortunately, this bit (intended to make the 0/1 optional) only works if there's no option after this one. Maybe check if the next option is 0/1/true/false and consume it as a value for KEEP_EXISTING_HOSTNAME only if that's the case?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -285,8 +292,8 @@ if test $KEEP_EXISTING_HOSTNAME -ne 1; then
   echo "$HOSTNAME" > /etc/HOSTNAME
 else
   HOSTNAME=$(cat /etc/HOSTNAME)
-  if ! echo "$HOSTNAME" | grep -q -o "$DOMAIN$"; then
-    echo "The fully qualified domain name did not contain the cloud domain."
+  if ! echo "$HOSTNAME" | grep -q -o "\.$DOMAIN$"; then
```
vuntz: Would be good to also check if `${HOSTNAME%%.$DOMAIN}` has no dot :-)

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
rsalevsky: Fixed

vuntz: I'm not sure that the correct fix, though :-) Should we really require this option to be the last?

(or am I misreading the code?)

rsalevsky: The second ;)

We require only a parameter for -k|--keep-existing-hostname where the option doesn't matter.

rhafer: > We require only a parameter for -k|--keep-existing-hostname where the option is doesn't matter.

How is that? Maybe I am also missreading the code. But if e.g. somebody calls ```crowbar_register -k -v``` (which is missing the 1/0 parameter for -k)  ```test $# -eq 1``` is false and ```KEEP_EXISTING_HOSTNAME``` will be set to ```-v```. I guess this is not what the code should do, right? Or am I missing something?

rsalevsky: Ah yet I understood what @vuntz said.

Your right, I changed the code a little bit to expect a number as parameter. But I think we should not over engineering the code. When `KEEP_EXISTING_HOSTNAME` is not `1` then the other code will not executed.

rhafer: +1 for the over-engineering :). Regarding that I am still wondering why we need -k to have a value at all. I guess the idea is to address https://github.com/crowbar/barclamp-provisioner/pull/398#discussion_r25850262 with this, right?

But do we really need/want to support the case where the user can override  ```node[:provisioner][:keep_existing_hostname]``` by calling the script with  -k 0? Maybe @vuntz can clarify. Otherwise I'd say do not expose the -k option at all when ```node[:provisioner][:keep_existing_hostname]`` is ```true```

vuntz: Agreed, my initial comment was to just not have this option in that case :-)

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -46,7 +50,9 @@ while test $# -gt 0; do
         -h|--help|--usage|-\?) usage ;;
         -v|--verbose) CROWBAR_VERBOSE=1 ;;
         -f|--force) CROWBAR_FORCE=1 ;;
+        <% if !node[:provisioner][:keep_existing_hostname] -%>
         -k|--keep-existing-hostname) KEEP_EXISTING_HOSTNAME=1 ;;
+        <%end%>
```
vuntz: Can you add spaces around `end`? :-) Also, use `-%>` to close the tag (and avoid a new line).

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -22,7 +22,11 @@
 
 usage () {
     cat <<EOF
+<% if !node[:provisioner][:keep_existing_hostname] -%>
 `basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [-k|--keep-existing-hostname] [--no-gpg-checks] [--interface IF]
+<% else %>
+`basename $0` [-h|--help] [-v|--verbose] [-f|--force] [--gpg-auto-import-keys] [--no-gpg-checks] [--interface IF]
+<% end %>
```
vuntz: Can you use `-%>` to close the else/end tags?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -359,8 +365,16 @@ fi
...
+MAC=$(cat /sys/class/net/$BOOTDEV/address)
 if test $KEEP_EXISTING_HOSTNAME -ne 1; then
+  HOSTNAME="d${MAC//:/-}.${DOMAIN}"
   echo "$HOSTNAME" > /etc/HOSTNAME
+else
+  HOSTNAME=$(cat /etc/HOSTNAME)
+  if ! echo "$HOSTNAME" | grep -q -o "\.$DOMAIN$" || echo "${HOSTNAME%%.$DOMAIN}" | grep -q -o "\."; then
+    echo "The fully qualified domain name did not contain the $DOMAIN cloud domain."
+    exit 1
+  fi
```
vuntz: Hrm, this code got moved earlier I think.

rsalevsky: I think you removed this mistakenly while rebasing #384.

vuntz: It's not removed, it's at https://github.com/rsalevsky/barclamp-provisioner/blob/keep_hostname_improvments/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb#L254

So you need to update the code there.

tboerger:
I would just skip the ```-k``` option but otherwise it LGTM

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -251,8 +257,8 @@ if test $KEEP_EXISTING_HOSTNAME -ne 1; then
   HOSTNAME="d${MAC//:/-}.${DOMAIN}"
 else
   HOSTNAME=$(cat /etc/HOSTNAME)
-  if ! echo "$HOSTNAME" | grep -q -o "$DOMAIN$"; then
-    echo "The fully qualified domain name did not contain the cloud domain."
+  if ! echo "$HOSTNAME" | grep -q -o "\.$DOMAIN$" || echo "${HOSTNAME%%.$DOMAIN}" | grep -q -o "\."; then
```
vuntz: (stupid question: why do we need `-o`?)

I'm not sure the error string below makes sense for the second test (`echo "${HOSTNAME%%.$DOMAIN}" | grep -q -o "\."`); I think I'd just do a `elif` with that test and a different string.

vuntz:
This looks good, just some minor comments.

rhafer:
Looks good to me. A somewhat more descriptive commit message (what are the improvements?) would be helpful though. But I am fine with merging this as is.

PR #404, opened 2015-03-05
Don't directly start dhcp server

toabctl:
The following error was introducted with 8f539ad0d83271c:

    FATAL: Mixlib::ShellOut::ShellCommandFailed:
    service[dhcp3-server] (dhcp::default line 154) had an error:
    Mixlib::ShellOut::ShellCommandFailed: Expe
    cted process to exit with [0], but received '1'
    ---- Begin output of /sbin/service dhcpd start ----
    STDOUT: Starting ISC DHCPv4 4.x Server
    please see /var/log/rc.dhcpd.log for details
    STDERR: ..failed
    ---- End output of /sbin/service dhcpd start ----
    Ran /sbin/service dhcpd start returned 1

Problem was that there was no subnet defined so the dhcp-server was
unable to start. The subnet file is now changed whenever "enable_pxe"
changes and this triggers a delayed "restart" notification for the
dhcp-server.

### DIFF
--- a/chef/cookbooks/dhcp/recipes/default.rb
+++ b/chef/cookbooks/dhcp/recipes/default.rb
@@ -164,6 +164,6 @@ service "dhcp3-server" do
     end
   end
   supports :restart => true, :status => true, :reload => true
-  action node[:provisioner][:enable_pxe] ? ["enable", "start"] : ["disable", "stop"]
+  action node[:provisioner][:enable_pxe] ? "enable" : ["disable", "stop"]
 end
 
--- a/chef/cookbooks/dhcp/templates/default/subnet.conf.erb
+++ b/chef/cookbooks/dhcp/templates/default/subnet.conf.erb
@@ -1,4 +1,5 @@
 # File managed by Crowbar
+<% if node[:provisioner][:enable_pxe] -%>
 
 subnet <%= @network["subnet"] %> netmask <%= @network["netmask"]%> {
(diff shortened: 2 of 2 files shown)


vuntz:
Might be good to mention that prior to the commit breaking this, we only had `:enable` too. So it used to work that way.

+1

toabctl:
> Might be good to mention that prior to the commit breaking this, we only had :enable too. So it used to work that way.

Yes. I changed it to ```[:enable, :start]``` because ```dhcp-server``` didn't start when you first disabled and then enabled the ```enable_pxe``` switch. In that case (given that nothing changed in between, i.e. no new nodes added) the start of the ```dhcp-server``` was not triggered by any configuration change. That's now solved with the big ```if``` in the template.

rsalevsky:
+1 works for me

vuntz:
> > Might be good to mention that prior to the commit breaking this, we only had :enable too. So it used to work that way.
> 
> Yes. I changed it to ```[:enable, :start]``` because ```dhcp-server``` didn't start when you first disabled and then enabled the ```enable_pxe``` switch. In that case (given that nothing changed in between, i.e. no new nodes added) the start of the ```dhcp-server``` was not triggered by any configuration change. That's now solved with the big ```if``` in the template.

Well, I guess it's still broken if you enable, disable, enable, disable :-) Maybe the file should be removed in a `else`?

PR #399, opened 2015-02-27
Migrate original autoyast repos

jsuchome:

PR #405, opened 2015-03-25
Fix 022_upgrade_autoyast_repos migration

rhafer:
The "suse" hierarchy in the proposal is optional and might not exist when
installing the barclamp.

### DIFF
--- a/chef/data_bags/crowbar/migrate/provisioner/022_upgrade_autoyast_repos.rb
+++ b/chef/data_bags/crowbar/migrate/provisioner/022_upgrade_autoyast_repos.rb
@@ -1,5 +1,7 @@
 def upgrade ta, td, a, d
-  if !a['suse']['autoyast']['repos'].nil? && !a['suse']['autoyast']['repos'].empty?
+  # The 'suse' hierarchy is optional in the proposal template so we need to
+  # check the whole tree here.
+  if a.fetch('suse', {}).fetch('autoyast',{})['repos']
     if %w(common suse-11.3 suse-12.0).select {|k|
         a['suse']['autoyast']['repos'].keys.include? k
     }.empty?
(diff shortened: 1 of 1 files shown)


rhafer:
This was introduced by: https://github.com/crowbar/barclamp-provisioner/pull/399

Note: I have not tested this.

chef/data_bags/crowbar/migrate/provisioner/022_upgrade_autoyast_repos.rb
```diff
@@ -12,10 +18,16 @@ def upgrade ta, td, a, d
...
-    a['suse'].delete('autoyast')
+  # The 'suse' hierarchy is optional in the proposal template
+  if !a['suse'].nil?
+    if !a['suse']['autoyast'].nil? &&
+          !a['suse']['autoyast']['repos'].nil? &&
+          !a['suse']['autoyast']['repos']['suse-11.3'].nil? &&
+          !a['suse']['autoyast']['repos']['suse-11.3'].empty?
+      a['suse']['autoyast']['repos'] = a['suse']['autoyast']['repos']['suse-11.3']
+    else
+      a.delete('suse')
```
vuntz: I would not delete the whole suse attribute (or is there nothing else in there?), just the autoyast bit.

jsuchome: We did not introduce the ``'suse'`` section with the previous commits, only the structure under ``['autoyast']['repos']`` was changed. ...

rhafer: > I would not delete the whole suse attribute (or is there nothing else in there?), just the autoyast bit.

There is nothing else in it AFAICS. At least the schema only allows it to contain the "autoyast" tree.

mjura:
LGTM

rhafer:
> Is a.fetch('suse', Hash.new).fetch('autoyast', Hash.new)['repos'] more idiomatic?

@aplanas I updated the pull request accordingly, is that what you meant? I must admit I am not really sure which way I like better :)

aplanas:
@rhafer Don't take my word when I talk about Ruby.  My proposal was mostly a question based on this:

http://stackoverflow.com/questions/1820451/ruby-style-how-to-check-whether-a-nested-hash-element-exists

Looks like that the .nil? and the empty? are avoidable too, and the condition can be written like this:

```ruby
if !a.fetch('suse', {}).fetch('autoyast', {})['repos']
  if %w(common suse-11.3 suse-12.0).select {
    |k| a['suse']['autoyast']['repos'].keys.include? k
  }.empty?
end
```

But is only a detail.

rhafer:
> Looks like that the .nil? and the empty? are avoidable too, and the condition can be written like this:
> 
> ```if !a.fetch('suse', {}).fetch('autoyast', {})['repos']```

I guess it would need to be then : ```if a.fetch('suse', {}).fetch('autoyast', {})['repos']``` (note the missing ```!```
But I agree that looks some what cleaner.

vuntz:
+1

aplanas:
+1

rsalevsky:
+1

PR #406, opened 2015-03-27
Make Windows/Hyper-V Admin-Password configurable

thutterer:
This change still needs some testing. **Do not merge yet!**
I open the PR for review and discussion:
- Should there be a helper text next to the password field that describes what characters are (not) allowed?
- Should the password be validated before it gets accepted?
- Is the number in the migration filename correct?

### DIFF
--- a/chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
+++ b/chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
@@ -1,0 +1,9 @@
+def upgrade ta, td, a, d
+  a['admin_pass'] = ta['admin_pass']
+  return a, d
+end
+
+def downgrade ta, td, a, d
+  a.delete('admin_pass')
+  return a, d
+end
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -278,7 +278,7 @@ if not nodes.nil? and not nodes.empty?
                       :admin_ip => admin_ip,
                       :admin_name => node[:hostname],
                       :crowbar_key => crowbar_key,
-                      :admin_pass => "crowbar",
+                      :admin_pass => node[:provisioner][:admin_pass],
                       :domain_name => node[:dns].nil? ? node[:domain] : (node[:dns][:domain] || node[:domain]))
           end
 
--- a/chef/data_bags/crowbar/bc-template-provisioner.json
+++ b/chef/data_bags/crowbar/bc-template-provisioner.json
@@ -76,6 +76,7 @@
       "use_local_security": true,
       "use_serial_console": false,
       "serial_tty": "ttyS0,115200n8",
+      "admin_pass": "crowbar",
       "dhcp": {
         "lease-time": 60,
         "state_machine": {
@@ -103,7 +104,7 @@
     "provisioner": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 22,
+      "schema-revision": 23,
       "element_states": {
         "provisioner-bootdisk-finder": [ "hardware-installing" ],
         "provisioner-server": [ "readying", "ready", "applying" ],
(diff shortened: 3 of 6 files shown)


nkrinner:
The migration file name could as well be `023_add_hyperv-adminpassword.rb`. In every case, the `schema-revision` in `bc-template-provisioner.json` has to be set to 23 (or 30)

vuntz:
> In every case, the schema-revision in bc-template-provisioner.json has to be set to 23 (or 30)

Target is tex, so it should be 23. We'll bump to 30 (or more) when we go for the next version.

tboerger:
I would prefer ```admin_password``` but otherwise it looks good. :+1:

thutterer:
> I would prefer admin_password

The name was `admin_pass` already before my changes, so I left it this way but I agree with you.
Any objections to change it to `admin_password`?

thutterer:
Tested and works! Can we merge this now?

chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade ta, td, a, d
+  a['admin_pass'] = ta['admin_pass']
+  return a, d
+end
+
+def downgrade ta, td, a, d
+  a.delete('admin_pass')
```
mjura: This shouldn't be deleted

ovisan:
There should be a check performed for the windows default password complexity, can be found here:
https://technet.microsoft.com/en-us/library/cc786468(v=ws.10).aspx

ovisan:
Also, implement a check to show an error if the password field is empty

chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
tboerger: Deleting it seems to be right as this attribute didn't exist before.

tboerger:
> Also, implement a check to show an error if the password field is empty

There should already be an error as this field is required.

> There should be a check performed for the windows default password complexity, can be found here:
> https://technet.microsoft.com/en-us/library/cc786468(v=ws.10).aspx
> 
> Note: only initial passwords (from unattended config file) can be simple, so a check is needed to be able to change the password later on.

As far as i remember we don't have real validation methods yet. The only option should be some javascript implementation in the barclamp specific javascript file at crowbar_framework/app/assets/javascripts/barclamps/provisioner/application.js

vuntz:
> As far as i remember we don't have real validation methods yet.

Hrm? We do have ` validate_proposal` and `validate_proposal_after_save`. Or am I misunderstanding?

rsalevsky:
Yes, but validate the password after sending is slightly annoying. I prefer also the javascript way but I think we should implement this in a more general way. Maybe we need them later also in other barclamps.

chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
mjura: @tboerger: yeap, you are right

mjura:
I agree that we need some password validation method.

nkrinner:
+1

@thutterer : please file a bug about password validation so it will get done at some point. But this is not within the scope of the card or this pull request afais.

ovisan:
If an existing password should be changed, then the validation must be performed, otherwise it will fail.

vuntz:
Argh, this got merged before I was able to add my comment -- I guess I was too slow ;-)

My main concern here is that it's not visible anywhere that this is windows-specific. IMHO, the string should mention it and the path to the attribute should also make it clear (`node[:provisioner][:windows][:admin_password]`).

vuntz:
Oh, and the string should make it clear that it sets the password at install-time only. If it gets changed later on, then we don't touch it.

tboerger:
Oh no, not merging... I talked to @thutterer yesterday and we agreed on the complexity validation on the serverside :(

jsuchome:
Sorry guys, it had already 3 +1's ... I assume the new issues could be handled by new PR

thutterer:
@tboerger There seem to be quite different opinions on how to approach the password validation. Others favor a javascript frontend solution. @nkrinner made the point that this discussion and any implementation is out of scope of this pull request and the related trello card. And as we only support to set the initial password for installation and not the changing of a password, the complexitiy check is not really needed.

@vuntz @jsuchome I started to work on a new PR where I will add Vincent's improvement ideas.

PR #408, opened 2015-04-23
Improve Windows/Hyper-V Admin-Password Feature

thutterer:
This is a follow-up on #406.
As discussed there, the attribute gets renamed to `admin_password` and put into a mapping called `windows` to clarify it's function.
In the Crowbar UI a hint gets added to explain it's functionality.

### DIFF
--- a/chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
+++ b/chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
@@ -1,9 +1,9 @@
 def upgrade ta, td, a, d
-  a['admin_pass'] = ta['admin_pass']
+  a['windows'] = ta['windows']
   return a, d
 end
 
 def downgrade ta, td, a, d
-  a.delete('admin_pass')
+  a.delete('windows')
   return a, d
 end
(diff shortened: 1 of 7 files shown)


chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
```diff
@@ -1,9 +1,10 @@
 def upgrade ta, td, a, d
-  a['admin_pass'] = ta['admin_pass']
+  a['windows'] = {}
+  a['windows']['admin_password'] = ta['admin_pass']
```
vuntz: You can simplify with `a['windows'] = ta['windows']`, but it's no big deal :-)

vuntz:
+1.

Note for others: I'm not concerned about changing the old migration instead of creating a new one as this migration just got in and really, it won't impact anyone.

chef/data_bags/crowbar/migrate/provisioner/023_add_hyperv-adminpassword.rb
thutterer: Updated it to have your line in the migration file.

nkrinner:
+1

PR #409, opened 2015-04-24
give network time to be started (bnc#927997)

bmwiedemann:
In https://bugzilla.suse.com/show_bug.cgi?id=927997
we found that Dell Poweredge C6105 systems
do not transfer traffic for 30 seconds
after the interface comes up, which is why wicked timed out before

Also there is no point in letting it wait for DHCPv6 responses.

These additional params are only needed on SLE12+ but dont hurt on others either

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -218,6 +218,7 @@ if not nodes.nil? and not nodes.empty?
 
         when os =~ /^(open)?suse/
           append << "install=#{install_url} autoyast=#{node_url}/autoyast.xml"
+          append << "ifcfg=dhcp4 netwait=60"
 
           Provisioner::Repositories.inspect_repos(node)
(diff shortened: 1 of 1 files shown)


aspiers:
Sounds reasonable to me, +1

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -218,6 +218,7 @@
 
         when os =~ /^(open)?suse/
           append << "install=#{install_url} autoyast=#{node_url}/autoyast.xml"
+          append << "ifcfg=dhcp4 netwait=60"
```
vuntz: I understand we do this for SLE12. This doesn't have any downside on SLE11 (ie, this still works fine there)?

bmwiedemann: exactly. the ifcfg option is new in SLE12, so is ignored in SLE11 and the netwait does not hurt too

PR #410, opened 2015-04-28
Fix useradd call for SLE11 (bsc#928931)

rhafer:
useradd on SLE11 doesn't know the --home-dir argument. It only allows -d, which
also works on SLE12.

Fixed: https://bugzilla.suse.com/show_bug.cgi?id=928931


vuntz:
+1

rsalevsky:
:+1:

PR #411, opened 2015-04-29
crowbar_register: Fix error message for hostname test

vuntz:
We check if there's a dot after removing the domain we expect; when
that's the case, then we're in a subdomain.


rhafer:
+1

rsalevsky:
:+1:

PR #412, opened 2015-05-05
Make SUSE Cloud repos optional to allow Storage-only deployments

tserong:
This makes the SUSE Cloud repos optional, in the same way that SLE HA
and Storage repos are already optional, i.e. everything except the base
OS is now optional.

If the Storage repos are present but Cloud repos are not, the autoyast
profile is tweaked to perform a Storage-only deployment, by installing
ses-release instead of suse-sle12-cloud-compute-release.

Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -19,6 +19,25 @@ class Provisioner
     class << self
       def suse_optional_repos(version, type)
         case type
+        when :cloud
+          case version
+          when "11.3"
+            %w(
+              SLE-Cloud
+              SLE-Cloud-PTF
+              SUSE-OpenStack-Cloud-SLE11-6-Pool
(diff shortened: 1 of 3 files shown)


tserong:
See also https://github.com/crowbar/crowbar/pull/2092 and https://github.com/crowbar/barclamp-crowbar/pull/1266

tboerger:
LGTM

rsalevsky:
:+1:

rsalevsky:
@tserong Please rebase.

tserong:
Don't merge this yet, I still need to test it since rebasing on master

tserong:
Tested SES-only deploy with rebased patch, works fine.

rsalevsky:
Still +1 from my side.

@tboerger Can you check again?

tboerger:
It still looks ok.

PR #414, opened 2015-05-06
Use Chef::DataBag instead of ProposalObject for barclamp info

bkutil:
Refs https://github.com/crowbar/barclamp-crowbar/pull/1269

### DIFF
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -76,7 +76,7 @@ class ProvisionerService < ServiceObject
         add_role_to_instance_and_node("provisioner", inst, name, db, role, "provisioner-server")
 
         # Reload the roles
-        db = db.reload
+        db.reload
         role = RoleObject.find_role_by_name "provisioner-config-#{inst}"
       end
 
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -66,7 +66,7 @@ class ProvisionerService < ServiceObject
     #
     if state == "discovered"
       @logger.debug("Provisioner transition: discovered state for #{name} for #{state}")
-      db = ProposalObject.find_proposal "provisioner", inst
+      db = Proposal.where(barclamp: "provisioner", name: inst).first
 
       #
       # Add the first node as the provisioner server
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -143,7 +143,7 @@ def transition(inst, name, state)
         else
           node_barclamp = node_role
         end
-        bc_databag = ProposalObject.find_data_bag_item("barclamps/#{node_barclamp}")
+        bc_databag = Chef::DataBag.load("barclamps/#{node_barclamp}") rescue nil
```
vuntz: That's not correct (and broken with the move away from data bags).

When we discussed it the other day, though, we found out that we can just drop this whole block (`if state == "installing"`) as no barclamp is using a `unsupported_platform` attribute.

bkutil: > That's not correct (and broken with the move away from data bags).

Broken as it should rescue to an array? Or in what sense?

vuntz: Broken as the data bag should not exist anymore anyway :-)

tboerger: Did i remember correctly that we don't provide the unsupported platform attribute anymore?

vuntz: @tboerger yes, hence my suggestion to simply drop the whole block :-)

bkutil: @vuntz just to be clear: so everything inside 'installing' if block can go away?

tboerger: > @tboerger yes, hence my suggestion to simply drop the whole block :-)

Makes sense
:+1:

vuntz: @bkutil yes

bkutil: > @bkutil yes

Dropped.

crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -76,7 +76,7 @@ def transition(inst, name, state)
         add_role_to_instance_and_node("provisioner", inst, name, db, role, "provisioner-server")
 
         # Reload the roles
-        db = ProposalObject.find_proposal "provisioner", inst
+        db = db.reload
```
vuntz: Just `db.reload`, maybe?

bkutil: > Just `db.reload`, maybe?

Yep, fixed.

PR #416, opened 2015-05-17
Fix apache config file to work with apache 2.4 or later

vuntz:
Note that NameVirtualHost is gone now: this is automatically detected.


tboerger:
+1

rsalevsky:
+1

PR #420, opened 2015-05-20
Add an option to install additional packages

thutterer:
Tested and works.
What is not yet covered in this pull request is the comment from @vuntz in [trello](https://trello.com/c/QLQyKYf7) about distinction of distributions. It was not part of the acceptance criteria, but still, maybe you want to propose a viable way to cover. In the UI and the schema.

And are you happy with the hint text? I kept it rather short.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -301,6 +301,11 @@ if node["platform"] == "suse" && !node.roles.include?("provisioner-server")
         %x{zypper --non-interactive addrepo --refresh #{attrs[:url]} #{name}}
       end
     end
+    # install additional packages
+    os = "#{node[:platform]}-#{node[:platform_version]}"
+    if node[:provisioner][:packages][os]
+      node[:provisioner][:packages][os].each { |p| package p }
+    end
   end
 end
 
--- a/chef/cookbooks/provisioner/recipes/setup_base_images.rb
+++ b/chef/cookbooks/provisioner/recipes/setup_base_images.rb
@@ -424,6 +424,8 @@ node[:provisioner][:supported_oses].each do |os,params|
     Provisioner::Repositories.inspect_repos(node)
     repos = Provisioner::Repositories.get_repos(node, "suse", target_platform_version)
 
+    packages = node[:provisioner][:packages][os] || []
+
     template "#{os_dir}/crowbar_register" do
       mode 0644
       owner "root"
@@ -437,6 +439,7 @@ node[:provisioner][:supported_oses].each do |os,params|
                 :crowbar_key => crowbar_key,
                 :domain => domain_name,
                 :repos => repos,
+                :packages => packages,
                 :target_platform_version => target_platform_version)
     end
 
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -232,6 +232,8 @@ if not nodes.nil? and not nodes.empty?
             end
           end
 
+          packages = node[:provisioner][:packages][os] || []
+
           template "#{node_cfg_dir}/autoyast.xml" do
             mode 0644
             source "autoyast.xml.erb"
@@ -240,6 +242,7 @@ if not nodes.nil? and not nodes.empty?
             variables(
                       :admin_node_ip => admin_ip,
                       :web_port => web_port,
+                      :packages => packages,
                       :repos => repos,
                       :rootpw_hash => node[:provisioner][:root_password_hash] || "",
                       :timezone => timezone,
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -278,6 +278,9 @@ sync
   <% end -%>
       <package>supportutils-plugin-susecloud</package>
 <% end -%>
+<% @packages.each do |package| -%>
+      <package><%= package %></package>
+<% end -%>
     </packages>
     <patterns config:type="list">
       <pattern>Minimal</pattern>
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -297,6 +297,11 @@ PACKAGES_INSTALL=
 # Obvious dependencies
 PACKAGES_INSTALL="$PACKAGES_INSTALL openssh"
 
+<% unless @packages.empty? -%>
+# Additional packages
+PACKAGES_INSTALL="$PACKAGES_INSTALL <%= @packages.join(' ') %>"
+<% end -%>
+
 # From autoyast profile
 PATTERNS_INSTALL="$PATTERNS_INSTALL Minimal base"
 PACKAGES_INSTALL="$PACKAGES_INSTALL netcat ruby2.1-rubygem-chef supportutils-plugin-susecloud"
--- a/chef/data_bags/crowbar/bc-template-provisioner.json
+++ b/chef/data_bags/crowbar/bc-template-provisioner.json
@@ -106,7 +106,7 @@
     "provisioner": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 23,
+      "schema-revision": 24,
       "element_states": {
         "provisioner-bootdisk-finder": [ "hardware-installing" ],
         "provisioner-server": [ "readying", "ready", "applying" ],
--- a/chef/data_bags/crowbar/bc-template-provisioner.schema
+++ b/chef/data_bags/crowbar/bc-template-provisioner.schema
@@ -73,6 +73,17 @@
                 }
               }
             },
+            "packages": {
+              "type": "map",
+              "required": false,
+              "mapping": {
+                =: {
+                  "type": "seq",
+                  "required": false,
+                  "sequence": [ { "type": "str" } ]
+                }
+              }
+            },
             "use_local_security": { "type": "bool", "required": true },
             "use_serial_console": { "type": "bool", "required": true },
             "serial_tty": { "type": "str",  "required": true },
--- a/chef/data_bags/crowbar/migrate/provisioner/024_add_autoyast_packages.rb
+++ b/chef/data_bags/crowbar/migrate/provisioner/024_add_autoyast_packages.rb
@@ -1,0 +1,9 @@
+def upgrade ta, td, a, d
+  a['packages'] ||= {}
+  return a, d
+end
+
+def downgrade ta, td, a, d
+  a.delete('packages')
+  return a, d
+end


rsalevsky:
Whats happen when a package is not available in any of the repositorys?

thutterer:
> Whats happen when a package is not available in any of the repositorys?

Then the installer tells so in an early stage and waits for the user to click a button.

rsalevsky:
I'm not sure if this really what we want. I would assume that the error message times out and the installation continues.

thutterer:
iirc there are ways to do this in autoyast files. but i'm not sure if enabling such an option would affect other installation issues where we might want a different behaviour. i'll do some research ...

thutterer:
I had a discussion with @nkrinner:
- When a package is missing or a typo, the installer should stop. Otherwise the admin wouldn't be aware that something went wrong.
- But it would be good to tell the user in the webUI about the missing package when the barclamp is applied.

So I will stop thinking about AutoYaST (btw [this](https://www.suse.com/documentation/sles11/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Reporting) would be it) and try to figure out how to report an error on barclamp apply.

thutterer:
For distinguishing between SLES11 und SLES12 we discussed to do it the simple way by just having two text fields. Any more sophisticated ideas? :)

tboerger:
Do we need this packages really while the autoyast installation? Or should it be enough to install these packages afterwards with the chef-client run?

jsuchome:
I think we want this in autoyast installation. For chef-client, based installations, we could always install packages from recipes.

chef/data_bags/crowbar/bc-template-provisioner.json
```diff
@@ -79,6 +79,11 @@
       "windows": {
         "admin_password": "crowbar"
       },
+      "suse": {
+        "autoyast": {
+          "packages": [ ]
+        }
+      },
```
jsuchome: Isn't this in conflict with potential ``suse["autoyast"]["repos"]`` content?

tboerger: Why should that conflict? We just realized that @aspiers did no migration script to create the autoyast namespace so that @thutterer have to do it in his migration script.

tboerger:
I would prefer another naming like ```attributes[:packages][:suse_11sp3]``` or if that works without yml quoting within the i18n file maybe ```attributes[:packages][:suse][:11sp3]```, than we have no issues with the i18n keys and it's more self explaining when we install the packages within a chef-client run afterwards (Possibility to update the package list even after autoyast installation).

jsuchome:
As with naming, it would be better consistent with values of ``[:autoyast][:repos]`` which are actually ``suse-11.3``, ``suse-12.0`` and ``common`` (common maybe doesn't make sense here0

thutterer:
Needs a rebase at the end but otherwise is now ready for review.

There is one thing that might be considered a conceptional flaw:
- According to our discussion, the check for wrong package names should happen when the barclamp is changed and applied.
- Therefore I implemented it in `base.rb` and raise an exception when `zypper search` can't find the package in the node's repos.
- But if you for example have a typo in a SLES12 package name but have no SLES12 node to which the barclamp could be applied to, the check does not run as `base.rb` is executed with chef-client on the nodes itself.

I just realized this while testing and don't have a better solution yet. Maybe you have?

jsuchome:
I'm not sure if the package check should be done by the barclamp, I think it would be left to autoyast.  And as you correctly mention - you might not have such node installed yet, so the correct check could not be done before the installation...

jsuchome:
Maybe my perspective is wrong, but I just wouldn't do the package check in the ruby code at all.

thutterer:
Rebased for easier testing.

nkrinner:
I prefer having the package availability check happening when the change in the barclamp is made. Else we will have a failure when nodes are allocated (or we allocate without the admin knowing that selected packages are not installed). This would make debugging much more complicated as far as I see it. So I think the autoyast installer is not the best place.

thutterer:
While I generally agree with @nkrinner that a check on barclamp change would be nice, I also have to admit that this is more problematic than I thought in the beginning:
- For the current approach's problem see my previous comment.
- If we would move that test to a `provisioner-server` recipe, we couldn't use `zypper search` anymore as that will find no 12.0 packages from a 11.3 crowbar. Furthermore searching in local repos needs that you add them first with `zypper ar` and delete them afterwards. That's not pretty at all.
- We could do a simple file search to check if there is a rpm with that name in one of the repos. But this also sounds a bit too vague.

So maybe we really have to fall back to @jsuchome's suggestion to not test this on barclamp change at all but let the installer complain when some package is missing.

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -301,6 +301,21 @@
...
+      repos.each do |name, repo|
+        if system("zypper search -r #{repo[:url]} --match-exact -t package #{package}")
+          package_exists = true
+          break
+        end
+      end
+      unless package_exists
+        raise "Missing additional package #{package} on #{node[:fqdn]}. Abort!"
+      end
+    end
```
vuntz: So I'm confused by this block: this will be executed on nodes (not the admin server); which more or less means that the node will already have been installed and that the autoyast bit will already have been used. So this comes too late.

Or am I missing the purpose of the block?

chef/data_bags/crowbar/bc-template-provisioner.json
```diff
@@ -79,6 +79,14 @@
       "windows": {
         "admin_password": "crowbar"
       },
+      "suse": {
+        "autoyast": {
+          "packages": {
+            "suse-11_3": [ ],
+            "suse-12_0": [ ]
```
vuntz: As @jsuchome mentioned earlier:
> As with naming, it would be better consistent with values of [:autoyast][:repos] which are actually suse-11.3, suse-12.0 and common (common maybe doesn't make sense here)

(agreed we don't need common).

So better to use the dot instead of underscore. And yes, it will be an issue for what @tboerger mentioned, but I don't think we'll show that in the UI.

vuntz: Ah, now I see you put it in the UI. Hrm.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -240,6 +240,7 @@
             variables(
                       :admin_node_ip => admin_ip,
                       :web_port => web_port,
+                      :packages => node[:provisioner][:suse][:autoyast][:packages]["suse-" + target_platform_version.sub(".", "_")],
```
vuntz: Maybe use `os..sub(".", "_")`? (Still unsure we want this underscore there, but let's discuss in the other thread)

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -278,6 +278,9 @@ sync
   <% end -%>
       <package>supportutils-plugin-susecloud</package>
 <% end -%>
+      <% @packages.each do |package| -%>
+        <package><%= package %></package>
```
vuntz: Indentation is wrong here, should be at the same level as `<package>supportutils-plugin-susecloud</package>` earlier on.

tboerger:
I'm fine if we skip the UI parts, do we really want to provide the UI?

chef/cookbooks/provisioner/recipes/base.rb
thutterer: Right. As mentioned in previous comments, this test will be dropped.

chef/data_bags/crowbar/bc-template-provisioner.json
thutterer: The dot was replaced with an underscore for a reason. I gave a very brief explanation in the commit message: The i18n library cannot handle names like `suse-11.3`. Best option was to replace it with an underscore.

thutterer: > Ah, now I see you put it in the UI. Hrm.

Sorry, I don't get your comment. What have I put in the User Interface?

thutterer:
@tboerger :
> I'm fine if we skip the UI parts, do we really want to provide the UI?

What's wrong about the UI part? Or, why wouldn't you want to provide one?

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
thutterer: True. I'll fix that.

vuntz:
@thutterer so in short we have to choose between two bad options:

 * use a consistent naming for the keys (`suse-11.3`, like we use for repos) and not be able to display the option in the UI
 * use an inconsistent (and possibly confusing) naming for the keys (`suse-11_3`) and have a UI for this

I'm not sold on the fact that we need a UI for this, so I'd lean towards the first option.

chef/data_bags/crowbar/bc-template-provisioner.schema
```diff
@@ -49,6 +49,22 @@
...
+                    "packages": {
+                      "type": "map",
+                      "required": false,
+                      "mapping": {
+                        "suse-11_3": {
+                          "type": "seq",
+                          "required": false,
+                          "sequence": [ { "type": "str" } ]
+                        },
+                        "suse-12_0": {
```
vuntz: Btw, we should definitely not hard-code the OS for which this work in the schema. See how it's done for the `repos` attribute a few lines after this.

thutterer:
@vuntz @tboerger 
I made the changes you suggested:
- removed the UI part
- removed hard-coded version names

Also tested. Adding `suse-11.3` or whatever-named sequences of packages to the _raw_ barclamp works fine.

chef/data_bags/crowbar/migrate/provisioner/024_add_autoyast_packages.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade ta, td, a, d
+  a['suse'] ||= {}
+  a['suse']['autoyast'] ||= {}
+  a['suse']['autoyast']['packages'] ||= {}
+  return a, d
+end
+
+def downgrade ta, td, a, d
+  a['suse']['autoyast'].delete('packages')
```
vuntz: This could fail as `a['suse']` might not event exist. You could probably do something like:
`a.fetch('suse', {}).fetch('autoyast', {}).delete('packages')`

vuntz:
I think this is going in the right direction.

You should also update the `crowbar_register` script to install the packages.

chef/data_bags/crowbar/migrate/provisioner/024_add_autoyast_packages.rb
tboerger: But this get's created in the upgrade migration?

tboerger:
But anyway i would suggest to name it ```attributes["packages"]``` instead of the deep nesting and add it to the chef-client run as well. That way we are able to add additional packages after the autoyast install as well.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -232,6 +232,17 @@
             end
           end
 
+          packages = [ ]
+          if node[:provisioner][:suse]
```
tboerger: Just stop this deep nesting, makes the code much simpler.

chef/data_bags/crowbar/migrate/provisioner/024_add_autoyast_packages.rb
vuntz: The attribute is not required, so could be dropped later on.

vuntz:
> But anyway i would suggest to name it attributes["packages"] instead of the deep nesting and add it to the chef-client run as well. That way we are able to add additional packages after the autoyast install as well.

While I agree about moving the attribute, I think the reason I agree is not that we can install packages after autoyast, but because there's no reason to have this SUSE-specific ;-)

So totally fine with me to move the attribute up.

tboerger:
> While I agree about moving the attribute, I think the reason I agree is not that we can install packages after autoyast, but because there's no reason to have this SUSE-specific ;-)

And what do you think about integrating it into the chef run? :)

thutterer:
So, instead of `[:suse][:autoyast][:packages]` it should be just `[:packages]` on the top-level?

tboerger:
> So, instead of [:suse][:autoyast][:packages] it should be just [:packages] on the top-level?

:+1:

vuntz:
> And what do you think about integrating it into the chef run? :)

Well, the reason we wanted that in autoyast is that the node might not be able to boot without the packages. I don't mind also adding this to regular chef runs, but I'm not sure I see a benefit. No objection against this, though.

nkrinner:
It would allow to install packages on all nodes in an easy way, at any time. I like the idea.

jsuchome:
But isn't it actually a different feature (worth different PR)?
I'm sorry for lengthening already long discussion, but maybe we could have two different lists: one for install time (autoyast), where we want to make sure correct packages get installed even before chef-client run, and the other list for possible later upgrade of the installed nodes with custom packages.

tboerger:
> I'm sorry for lengthening already long discussion, but maybe we could have two different lists: one for install time (autoyast), where we want to make sure correct packages get installed even before chef-client run, and the other list for possible later upgrade of the installed nodes with custom packages.

Does this really matter or make sense to maintain different lists of packages?

thutterer:
Ok, I've updated and squashed the PR. Everything was successfully tested. Please give it a try!

chef/cookbooks/provisioner/recipes/base.rb
```diff
@@ -301,6 +301,13 @@
         %x{zypper --non-interactive addrepo --refresh #{attrs[:url]} #{name}}
       end
     end
+    # install additional packages
+    os = "#{node[:platform]}-#{node[:platform_version]}"
+    packages = [ ]
+    if node[:provisioner][:packages][os]
+      packages = node[:provisioner][:packages][os]
+    end
+    system("zypper --non-interactive install #{packages.join(' ')}")
```
vuntz: Any reason for not using the `package` chef resource?

thutterer: No, good point. You mean like this:
```ruby
    if node[:provisioner][:packages][os]
      node[:provisioner][:packages][os].each do |p|
        package p do
          action :install
        end
      end
    end
```
This also has the advantage that `base.rb` throws an error if one of the packages couldn't be installed. With my current approach, `zypper` does not really complain about missing packages.

tboerger: IMHO it have to complain about missing packages.

thutterer: > IMHO it have to complain about missing packages.

Yes, and with the suggestion from @vuntz it would.

Style question: Do we prefer curly braces for one-line blocks? Like:
```ruby
package p { action :install }
```
or even
```ruby
node[:provisioner][:packages][os].each { |p| package p { action :install } }
```

vuntz: You can simplify to:

```ruby
node[:provisioner][:packages][os].each { |p| package p }
```

thutterer: True as `:install` is default. Updated PR.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -240,6 +245,7 @@
             variables(
                       :admin_node_ip => admin_ip,
                       :web_port => web_port,
+                      :packages => packages,
```
tboerger: Why not simply ```node[:provisioner][:packages][os] || []``` like below for the ```:root_password_hash```?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -297,6 +297,11 @@ PACKAGES_INSTALL=
 # Obvious dependencies
 PACKAGES_INSTALL="$PACKAGES_INSTALL openssh"
 
+# Additional packages
+<% if node[:provisioner][:packages][@os] -%>
```
tboerger: This is ugly, assign this variables the same way as for the autoyast template.

chef/cookbooks/provisioner/recipes/update_nodes.rb
tboerger: Forget this and write ```packages = node[:provisioner][:packages][os] || []``` to assign it to the template resource of crowbar_register as well.

tboerger:
Some smaller changes but beside that :+1:

thutterer:
I followed @tboerger's suggestions and updated the PR again. Also tested that nothing broke.

jsuchome:
It is minor detail (and I don't know if others would agree), but I'd rename your ``packages`` to something like ``additional-packages`` so there's a distinction between what we install by default

thutterer:
@jsuchome 
> It is minor detail (and I don't know if others would agree), but I'd rename your packages to something like additional-packages so there's a distinction between what we install by default

Do you mean the variable names that are prepared for and used in `crowbar_register` and `autoyast.xml` or do you also mean the name of the chef attribute?

jsuchome:
Probably all of them to keep it consistent. Unless there's something I do not see that should prevent it...

nkrinner:
I think there are reasons for both `packages` and `additional-packages`, but it doesn't matter much for me. So I am happy with the current state.

+1

tboerger:
:+1:

thutterer:
Thanks to @rsalevsky I just realized that this does not work anymore when installing crowbar from scratch.
During refactoring I accidently changed this line so that it depends on `[:packages]` to exist which it does not initially (`"required": false` and no default set)

Creating a follow-up ... (and will add link to it here as soon as it's pushed)

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -424,6 +424,8 @@
     Provisioner::Repositories.inspect_repos(node)
     repos = Provisioner::Repositories.get_repos(node, "suse", target_platform_version)
 
+    packages = node[:provisioner][:packages][os] || []
```
thutterer: Thanks to @rsalevsky I just realized that this does not work anymore when installing crowbar from scratch.
During refactoring I accidently changed this line so that it depends on `[:packages]` to exist which it does not initially (`"required": false` and no default set)

Creating a follow-up ... (and will add link to it here as soon as it's pushed)

PR #422, opened 2015-06-03
Install SES only on SLE 12

rsalevsky:
SES is not available for older versions then SLE 12

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -268,7 +268,7 @@ sync
       <package>biosdevname</package>
       <package>netcat</package>
       <package>ruby2.1-rubygem-chef</package>
-<% if @is_ses -%>
+<% if @is_ses and @target_platform_version == "12.0" -%>
       <package>ses-release</package>
 <% else -%>
   <% if @target_platform_version == "11.3" -%>


chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -268,7 +268,7 @@ sync
       <package>biosdevname</package>
       <package>netcat</package>
       <package>ruby2.1-rubygem-chef</package>
-<% if @is_ses -%>
+<% if @is_ses and @target_platform_version == "12.0" -%>
```
vuntz: Can we have `@is_ses` true on other versions? I thought not?

rsalevsky: I'm not sure if this breaks something else this was my first idea to fix the issue. Maybe @tserong has a better one.

tserong: I must be missing something. What's the underlying problem? In this context `@is_ses` should only be true if the cloud repos are missing but storage repos are available.

vuntz: That was the code I was thinking of: https://github.com/crowbar/barclamp-provisioner/blob/master/chef/cookbooks/provisioner/recipes/update_nodes.rb#L253

And indeed, there's no check for the version there.

@tserong I guess the "problem" is that it will more or less always work, but in theory, the test is not perfect ;-)

vuntz:
So I'm +0.5 here. I would possibly do that in the recipe instead, but that doesn't really matter much (and might read cleaner the way you did it anyway).

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
tserong: Gotcha.

tserong:
I'm OK with this, although later once SP1 comes out it'll need tweaking (as will a couple of "suse-12.0" strings in barclamp-crowbar, for that matter)

dirkmueller:
IMHO we should rather fix the place in provisioner where is_ses is initialized (even in the case of sle11)

rsalevsky:
Continued in #430

PR #423, opened 2015-06-03
Fix Cloud repo names

rsalevsky:
The repos are named Cloud and Cloud-PTF without the SLE-. Let's use the
correct name instead of changin it on runtime. This is basicaly needed
because the gsub is missing on some other places.

### DIFF
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -23,15 +23,15 @@ class Provisioner
           case version
           when "11.3"
             %w(
-              SLE-Cloud
-              SLE-Cloud-PTF
+              Cloud
+              Cloud-PTF
               SUSE-OpenStack-Cloud-SLE11-6-Pool
               SUSE-OpenStack-Cloud-SLE11-6-Updates
             )
           when "12.0"
             %w(
-              SLE-Cloud
-              SLE-Cloud-PTF
+              Cloud
+              Cloud-PTF
               SUSE-OpenStack-Cloud-6-Pool
               SUSE-OpenStack-Cloud-6-Updates
             )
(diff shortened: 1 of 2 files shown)


vuntz:
Hrm. Where are the gsub bits missing?

In general, I agree with the changes, but was a bit reluctant to do the changes since I didn't know what would break :-)

rsalevsky:
For example here (https://github.com/crowbar/barclamp-provisioner/blob/master/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb#L32) and in some other files where the repo name is used. 

The Problem is we can be never sure to catch all lines where the gsub is needed. When I understood the code correct then this is just a name which should match the repo name in `/srv/tftpboot/suse-*/repos/`.

vuntz:
@rsalevsky I think it was on purpose to have the SLE- prefix in the autoyast profile.

But in general, I'm fine trying to get rid of that.

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -92,9 +92,8 @@ def inspect_repos(node)
             suse_optional_repos(version, :cloud).each do |name|
               repos[name] ||= Mash.new
               next unless repos[name][:url].nil?
-              path_name = name.sub(/^SLE-Cloud/, 'Cloud')
```
vuntz: There's another similar gsub later in the code, can you drop it too?

rsalevsky: Done

vuntz:
One more change: in crowbar_register, we have `zypper -n --gpg-auto-import-keys refresh -r SLE-Cloud-PTF`.

Then you will also need to update install-chef-suse.sh in the crowbar repo, which also mentions SLE-Cloud.

vuntz:
So let's see if it works, +1

dirkmueller:
I'm for getting rid of the SLE- prefix (we might want to discuss renaming it to OpenStack-Cloud though ;-))). +1

PR #424, opened 2015-06-08
read also mounted disks that are unclaimed

jsuchome:
When looking for unclaimed disks, explicitely ask
also for the mounted ones (that would be filtered otherwise)

Depends on https://github.com/crowbar/barclamp-deployer/pull/244


vuntz:
+1

tboerger:
:+1:

PR #425, opened 2015-06-10
Fix existence checks of packages lists for "Additional Packages" feature

thutterer:
This is a follow-up pull request to #420 
Refactoring introduced a bug that slipped through my tests.

The way I checked the existence of additional packages entries at the end raised an `undefined method '[]' for nil:NilClass` as long `:packages` did not exist yet.

### DIFF
--- a/chef/cookbooks/provisioner/recipes/setup_base_images.rb
+++ b/chef/cookbooks/provisioner/recipes/setup_base_images.rb
@@ -424,7 +424,10 @@ node[:provisioner][:supported_oses].each do |os,params|
     Provisioner::Repositories.inspect_repos(node)
     repos = Provisioner::Repositories.get_repos(node, "suse", target_platform_version)
 
-    packages = node[:provisioner][:packages][os] || []
+    packages = []
+    if node[:provisioner][:packages]
+      packages = node[:provisioner][:packages][os] || []
+    end
 
     template "#{os_dir}/crowbar_register" do
       mode 0644
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -232,7 +232,10 @@ if not nodes.nil? and not nodes.empty?
             end
           end
 
-          packages = node[:provisioner][:packages][os] || []
+          packages = []
+          if node[:provisioner][:packages]
+            packages = node[:provisioner][:packages][os] || []
+          end
 
           template "#{node_cfg_dir}/autoyast.xml" do
             mode 0644
--- a/chef/cookbooks/provisioner/recipes/base.rb
+++ b/chef/cookbooks/provisioner/recipes/base.rb
@@ -303,8 +303,10 @@ if node["platform"] == "suse" && !node.roles.include?("provisioner-server")
     end
     # install additional packages
     os = "#{node[:platform]}-#{node[:platform_version]}"
-    if node[:provisioner][:packages][os]
-      node[:provisioner][:packages][os].each { |p| package p }
+    if node[:provisioner][:packages]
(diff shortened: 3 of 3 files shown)


chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -424,7 +424,11 @@
     Provisioner::Repositories.inspect_repos(node)
     repos = Provisioner::Repositories.get_repos(node, "suse", target_platform_version)
 
-    packages = node[:provisioner][:packages][os] || []
+    if node[:provisioner][:packages]
+      packages = node[:provisioner][:packages][os] || []
+    else
+      packages = []
+    end
```
rsalevsky: I prefer the following style but I'am also fine with leaving it as it is.
```
packages = []
if node[:provisioner][:packages]
  packages = node[:provisioner][:packages][os] || []
end
```

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -232,7 +232,11 @@
             end
           end
 
-          packages = node[:provisioner][:packages][os] || []
+          if node[:provisioner][:packages]
+            packages = node[:provisioner][:packages][os] || []
+          else
+            packages = []
+          end
```
rsalevsky: Same again.

rsalevsky:
Beside my small nitpick I'm fine, crowbar installation works again. :+1:

chef/cookbooks/provisioner/recipes/setup_base_images.rb
thutterer: Actually I had your version before the evil rewrite. But I tend to favor my current style just for readability reasons. `If`/`else` seems to reflect the logic better than overwriting a set value in a simple `if`. But I guess this is a matter of taste and I'm happy to change it if someone else outvotes me.

thutterer:
@rsalevsky I updated according to your suggestion.

tboerger:
But you made a migration that created the packages key? so why do we need all the checks?

chef/cookbooks/provisioner/recipes/setup_base_images.rb
tboerger: And i would prefer

```
packages = if node[:provisioner][:packages]
  node[:provisioner][:packages][os] || []
else
  []
end
```

rsalevsky:
The packages are not required and we have no default values for them. So they are missing in an new environment.

chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -424,7 +424,10 @@
     Provisioner::Repositories.inspect_repos(node)
     repos = Provisioner::Repositories.get_repos(node, "suse", target_platform_version)
 
-    packages = node[:provisioner][:packages][os] || []
+    packages = []
+    if node[:provisioner][:packages]
+      packages = node[:provisioner][:packages][os] || []
+    end
```
thutterer: >And i would prefer
```
packages = if node[:provisioner][:packages]
  node[:provisioner][:packages][os] || []
else
  []
end
```

No strong opinion on my side. Others?

thutterer:
In an offline discussion @tboerger suggested to loose the code changes in this PR and instead make `:packages` a required attribute and set it's default value to `[ ]`.

dirkmueller:
and how would that make [packages][os] a required attribute ? that change would only solve one level of if's

thutterer:
I think #426 is a cleaner way to fix this. That way, `packages` attribute is always in the JSON and we can depend on it in the code without multiple conditions.

rsalevsky:
#426 is merged

PR #426, opened 2015-06-11
Fix "packages" attribute requirement

thutterer:
This is a follow-up pull request to #420 and an alternative to #425.
Refactoring introduced a bug that slipped through my tests.

The `packages` attribute was not required in the schema and got only created in the migration but not in the initial JSON. So Crowbar installation failed.

### DIFF
--- a/chef/data_bags/crowbar/bc-template-provisioner.json
+++ b/chef/data_bags/crowbar/bc-template-provisioner.json
@@ -73,6 +73,7 @@
       "timezone": "UTC",
       "web_port": 8091,
       "coredump": false,
+      "packages": {},
       "use_local_security": true,
       "use_serial_console": false,
       "serial_tty": "ttyS0,115200n8",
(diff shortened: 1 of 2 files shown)


jsuchome:
Does it really need to be required?

thutterer:
> Does it really need to be required?

Well, my understanding is, that if the code depends on it to exist, then it is indeed required.
Of course we could also use the extra conditions from my first pull request and set its `require` back to `false`.
But @tboerger made the point yesterday that creating `packages` as an empy `[ ]` initially has the advantage that this feature is a bit more obvious to use and also less error-prone because the attribute is already in the right place.

dirkmueller:
you need to add a migration then creating the module if it is missing

tboerger:
> you need to add a migration then creating the module if it is missing

The migration is already merged: https://github.com/crowbar/barclamp-provisioner/blob/master/chef/data_bags/crowbar/migrate/provisioner/024_add_autoyast_packages.rb

tboerger:
:+1:

dirkmueller:
ah, ok. +1

PR #427, opened 2015-06-12
Allow only legal characters in names of additional packages and put them in quotes

thutterer:
I would like to improve string handling in the _additional packages_ feature.
Without that, it accepts any string as a package name.
This can be exploited with `crowbar_register` e.g. with `; rm foo; zypper in ` as a package name.

The first commit adds a validation RegEx that only allows characters from the [package naming guidelines](https://en.opensuse.org/openSUSE:Package_naming_guidelines#Common_Character_Set_for_Package_Naming) plus the version number operater chars (`< > =`).

The second commit concerns only `crowbar_register` and puts quotes around every package name there. Otherwise you couldn't use the version operator as you would redirect stdin or stdout. (e.g. `'emacs>4.0'`)

### DIFF
--- a/crowbar_framework/app/models/provisioner_service.rb
+++ b/crowbar_framework/app/models/provisioner_service.rb
@@ -51,6 +51,14 @@ class ProvisionerService < ServiceObject
   end
 
   def validate_proposal_after_save proposal
+    proposal["attributes"]["provisioner"]["packages"].each do |platform, packages|
+      packages.each do |package|
+        unless Crowbar::Validator::PackageNameValidator.new.validate(package)
+          validation_error("Package \"#{package}\" for \"#{platform}\" did not pass name validation.")
+        end
+      end
+    end
+
     validate_one_for_role proposal, "provisioner-server"
 
     super
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -297,11 +297,6 @@ PACKAGES_INSTALL=
 # Obvious dependencies
 PACKAGES_INSTALL="$PACKAGES_INSTALL openssh"
 
-<% unless @packages.empty? -%>
-# Additional packages
-PACKAGES_INSTALL="$PACKAGES_INSTALL <%= @packages.join(' ') %>"
-<% end -%>
-
 # From autoyast profile
 PATTERNS_INSTALL="$PATTERNS_INSTALL Minimal base"
 PACKAGES_INSTALL="$PACKAGES_INSTALL netcat ruby2.1-rubygem-chef supportutils-plugin-susecloud"
@@ -317,7 +312,7 @@ zypper --non-interactive install --auto-agree-with-licenses suse-sle11-openstack
 <% else -%>
 zypper --non-interactive install --auto-agree-with-licenses suse-openstack-cloud-release
 <% end -%>
-zypper --non-interactive install $PACKAGES_INSTALL
+zypper --non-interactive install $PACKAGES_INSTALL<%= " '#{@packages.join("' '")}'" unless @packages.empty? %>
 
 # Fail early if we know we can't succeed because of a missing chef-client
 if ! which chef-client &> /dev/null; then


crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -51,6 +51,16 @@ def create_proposal
   end
 
   def validate_proposal_after_save proposal
+    proposal["attributes"]["provisioner"]["packages"].each do |packages_for_platform|
```
tboerger: proposal["attributes"]["provisioner"]["packages"].each do |platform, packages|

crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -51,6 +51,16 @@ def create_proposal
...
+      platform_version = packages_for_platform[0]
+      platform_packages = packages_for_platform[1]
+      platform_packages.each do |package|
+        unless package =~ /(^\w[\w\+\.-<=>]+$)/
```
tboerger: You missed an underscore, and put the dash at the end of the regex.

thutterer: Underscore is a `\w` [word character](http://ruby-doc.org/core-2.1.1/Regexp.html#class-Regexp-label-Character+Classes).
The position of the dot doesn't matter. It's meant as the character `.` and not as `/./` ... or do you ask for this just for cosmetic reasons?

tboerger: > Underscore is a \w word character.

Yes, you are right, sorry.

> The position of the dot doesn't matter. It's meant as the character \. and not as /./ ... or do you ask for this just for cosmetic reasons?

I'm not talking about the dot, just about the dash. It is mostly safer to put it at the end to avoid some bad behavior like taking the dash for a rang instead of matching for a dash :)

thutterer: > I'm not talking about the dot, just about the dash.

Ups, misread. I'm not a morning person :) Thanks for the tip. I'll move it.

crowbar_framework/app/models/provisioner_service.rb
thutterer: Right! Forgot about that shortcut. I changed the commit accordingly.

crowbar_framework/app/models/provisioner_service.rb
```diff
@@ -51,6 +51,14 @@ def create_proposal
   end
 
   def validate_proposal_after_save proposal
+    proposal["attributes"]["provisioner"]["packages"].each do |platform, packages|
+      packages.each do |package|
+        unless package =~ /(^\w[\w\+\.<=>-]+$)/
```
thutterer: What do you think of this instead?
`(^\w[\w\+\.-]?[<>]?=?[\w\+\.-]+$)`
This prevents wrong usage of the version operators. Only downside I see is that package names now must consist of at least two characters.

tboerger: I'm not an regex specialist, but LGTM. There are anyway no packages below 2 chars, so that sounds ok to me.

jsuchome:
Good to use Package naming guideline, but are we sure all our packages actually pass this guideline?

thutterer:
> Good to use Package naming guideline, but are we sure all our packages actually pass this guideline?

I wouldn't know any packages that don't. I mean they have to pass those guidelines to get into the BuildService, right?

rsalevsky:
Not really, you can not expect that the packages comes from the build service. Customers can have there own repositories and we support also other platforms then SUSE.

jsuchome:
ok

thutterer:
@rsalevsky 
> you can not expect that the packages comes from the build service. Customers can have there own repositories

So, what are you suggesting? Allow more characters? Or change the RegEx so that only the evil characters are disallowed? Loose the validation completly?

> and we support also other platforms then SUSE

Alle three ways of installing additional packages are only supported with SUSE systems.

rsalevsky:
I'm not sure whats the best way to handle that. Maybe we should only disallow some characters instead of allowing them.

thutterer:
Ok, then it would be either like ```\A[^;&\n'"`]+\z``` if we only want to disallow a few characters.
Or something like ```\A([^;&\n'"`<>=]+(?:[<>]?=?[^;&\n'"`<>=]+)?)\z``` if we want to keep the rules about version operators.

jsuchome:
In the end, we could force customers to use some reasonable naming for their packages, right? They could even change it when they get the validation error, so it might not be bad idea.

thutterer:
The suggested black-list Regular Expressions in my previous comment don't exclude `\s` (any whitespace character). Probably we should add it...
Or go back to my initial white-list RegEx?

jsuchome:
I'm fine with white list actually. +1 (haven't tested it myself)

nkrinner:
+1, but not tested. I am fine with the regex as it allows what I consider being an industry standard.

tboerger:
Currently I would prefer the current regex.

rsalevsky:
+1

aspiers:
This discussion reminds me a bit of section 5 of "[What is Code?](http://www.bloomberg.com/graphics/2015-paul-ford-what-is-code/)" ;-)

-1 from me (sorry).  A better approach would be to extract the regexp validation code into a helper library (BTW we already have at least one of those in Crowbar) and then write unit tests which prove the regexp is correct.  Without those tests, there is a very good chance the regexp will be wrong, or will easily break in the future if someone changes it.  Who knows, maybe we already have another package validation regexp elsewhere in Crowbar ...

aspiers:
See https://github.com/crowbar/barclamp-crowbar/issues/192 and https://github.com/crowbar/crowbar/issues/1619.  It looks like validation is currently a bit of a mess:

- [ ] [the `CrowbarValidator` class](https://github.com/crowbar/barclamp-crowbar/blob/master/crowbar_framework/app/models/crowbar_validator.rb) is not a model so should not live in `app/models`
- [ ] no proper unit test coverage of `CrowbarValidator`

I think I would suggest putting the package validation regexp in a new `crowbar_framework/lib/validation.rb` or similar inside the `crowbar` barclamp, and then corresponding unit tests in `crowbar_framework/spec/lib/validation.rb` or similar.  Moving already existing validation code to the correct place can happen separately at another time.

aspiers:
Just to clarify the thinking behind my -1 ... quite a while ago, the team agreed on [a Definition Of Done](https://github.com/SUSE/cloud/blob/master/docs/Scrum/scrum-definitions.md) which includes the text:

* tests written and passing to cover any changes in supported behaviour, and no test breakage introduced, unless PO explicitly grants an exception
    * This needs some discussion when talking about barclamp tests since, for most barclamps, there is some initial setup that we should take care of.

To me this means that if it's reasonably easy to write tests, we should, but if it's so hard that it would require a huge amount of work (e.g. due to requirement for heavy mocking/stubbing), then maybe we can skip the requirement and instead rely on integration tests (i.e. Jenkins).

However in this case, it's really very easy indeed to write RSpec unit tests to test a single regular expression, since it's just a question of putting a single function in a dependency-free library and testing that in isolation.  Writing tests doesn't get any easier than this situation!  So if we don't even try to do it here (as part of the sprint), we're totally failing to aim for the bar we set ourselves when we adopted Scrum, and IMHO that helps set a very bad precedent.

tboerger:
How do you get to CrowbarValidator? This is a totally different story and a mess as well. Overall the current validation process is a mess and needs a much bigger refactoring. IMHO the -1 is not really valid...

aspiers:
@tboerger Please read what I wrote more closely.  I already said that moving `CrowbarValidator` should be done separately.  However they are both validation code, which is why I mentioned it here.

Also I already gave a detailed justification for the -1.  If you wish to dispute that, please do us the courtesy of providing a counter-argument, rather than just an "IMHO".

tboerger:
I already gave an explaination... 

> Overall the current validation process is a mess and needs a much bigger refactoring.

aspiers:
That's not an explanation, that's just repeating [something I already said](https://github.com/crowbar/barclamp-provisioner/pull/427#issuecomment-113083886), which is totally independent of this PR (even though it is partially related to it).

The -1 was specifically due to the lack of tests for the regexp in this PR, so it has nothing to do with the bigger validation mess.  And anyway, @thutterer has already very kindly done the right thing and submitted crowbar/barclamp-crowbar#1290, so as soon as this PR is switched to use that new library, I can change my -1 to a +1.

thutterer:
Ok, PR is now updated to use the validation module from crowbar/barclamp-crowbar#1290.
If my counting is correct, it now has +4 :)
But please consider that this must not be merged before crowbar/barclamp-crowbar#1290 got merged!

tboerger:
:+1:

rsalevsky:
:+1:

PR #428, opened 2015-06-15
Refactor missing_* to *_available

rsalevsky:
Logically this behavior makes more sense instead of saying what is missing.

Depends on:
https://github.com/crowbar/barclamp-pacemaker/pull/199
https://github.com/crowbar/barclamp-ceph/pull/162

### DIFF
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -81,9 +81,9 @@ class Provisioner
         case node[:platform]
         when "suse"
           repos = Mash.new
-          missing_cloud = false
-          missing_hae = false
-          missing_storage = false
+          cloud_available = false
+          hae_available = false
+          storage_available = false
 
           %w(11.3 12.0).each do |version|
             repos.merge! suse_get_repos_from_attributes(node,"suse",version)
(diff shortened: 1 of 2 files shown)


chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -92,40 +92,41 @@ def inspect_repos(node)
...
-          # know that HA cannot be used
-          # know that SUSE_Storage cannot be used
+          # set an attribute about installed repos so that cookbooks and crowbar
+          # know that HA can be used
+          # know that SUSE_Storage can be used
+          # know that OpenStack can be used
           node_set = false
           node.set[:provisioner][:suse] ||= {}
-          if node[:provisioner][:suse][:missing_cloud] != missing_cloud
-            node.set[:provisioner][:suse][:missing_cloud] = missing_cloud
```
toabctl: I think we need a migration for the keys `:missing_cloud`, `:missing_hae` and `:missing_storage`.

tboerger: No, this is nothing we want to migrate because these values are set on the fly to detect if we are missing some repositories.

tboerger: But as a side note... You have to check that this attribute gets referenced in the other barclamps as well e.g. pacemaker.

rsalevsky: I greped over the whole crowbar organisation and found only a reference in barcalamp-ceph and barclamp-pacemaker.

tboerger: > I greped over the whole crowbar organisation and found only a reference in barcalamp-ceph and barclamp-pacemaker.

:+1:

mjura:
+1

aspiers:
-1 ...  While I give +1 for the idea of negating the logic, `install_*` is a misnomer.  I think it should be `*_available`; just because the HAE repos are available, it doesn't mean you will install anything from them.  There is a fundamental difference.  So if you change `install_*` to `*_available` then it gets my +1.

rsalevsky:
@aspiers Can you please review again?

chef/cookbooks/provisioner/libraries/repositories.rb
```diff
@@ -92,40 +92,41 @@ def inspect_repos(node)
...
               next unless repos[name][:url].nil?
-              missing_storage ||= !(File.exists? "#{node[:provisioner][:root]}/suse-#{version}/repos/#{name}/repodata/repomd.xml")
+              storage_available ||= File.exist?("#{node[:provisioner][:root]}/suse-#{version}/repos/#{name}/repodata/repomd.xml")
             end
           end
 
-          # set an attribute about missing repos so that cookbooks and crowbar
-          # know that HA cannot be used
-          # know that SUSE_Storage cannot be used
+          # set an attribute about installed repos so that cookbooks and crowbar
```
aspiers: "available repos" not "installed repos" :-)

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -254,7 +254,7 @@
                       :node_hostname => mnode[:hostname],
                       :target_platform_version => target_platform_version,
                       :is_ses => node[:provisioner][:suse] &&
-                        node[:provisioner][:suse][:missing_cloud] && !node[:provisioner][:suse][:missing_storage],
+                        !node[:provisioner][:suse][:cloud_available] && node[:provisioner][:suse][:storage_availabl],
```
aspiers: `:storage_availabl` typo

aspiers:
Still 2 bugs, but +1 when those are fixed.

aspiers:
Thanks, LGTM now.

PR #430, opened 2015-06-23
Refactor package selection and installation

rsalevsky:
- install cloud packages only when the cloud repo is available
- install SES only on SLE12

Superseeds #422

### DIFF
--- a/chef/cookbooks/provisioner/recipes/update_nodes.rb
+++ b/chef/cookbooks/provisioner/recipes/update_nodes.rb
@@ -225,14 +225,32 @@ if not nodes.nil? and not nodes.empty?
           repos = Provisioner::Repositories.get_repos(node, "suse", target_platform_version)
           Chef::Log.info("repos: #{repos.inspect}")
 
+          packages = node[:provisioner][:packages][os] || []
+          packages.push("autoyast2-installation")
+          packages.push("biosdevname")
+          packages.push("netcat")
+          packages.push("ruby2.1-rubygem-chef")
+
           if node[:provisioner][:suse]
             if node[:provisioner][:suse][:autoyast]
               ssh_password = node[:provisioner][:suse][:autoyast][:ssh_password]
               append << "UseSSH=1 SSHPassword=#{ssh_password}" if ssh_password
             end
-          end
 
-          packages = node[:provisioner][:packages][os] || []
+            if node[:provisioner][:suse][:cloud_available]
+              packages.push("supportutils-plugin-susecloud")
+
+              if target_platform_version == "11.3"
+                packages.push("use-sle11-openstack-cloud-release")
+              else
+                packages.push("suse-openstack-cloud-release")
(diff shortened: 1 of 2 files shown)


tboerger:
I'm missing an explanation why this change is needed

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -225,14 +225,32 @@
...
               append << "UseSSH=1 SSHPassword=#{ssh_password}" if ssh_password
             end
-          end
 
-          packages = node[:provisioner][:packages][os] || []
+            if node[:provisioner][:suse][:cloud_available]
+              packages.push("supportutils-plugin-susecloud")
+
+              if target_platform_version == "11.3"
+                packages.push("use-sle11-openstack-cloud-release")
```
vuntz: Typo: use => suse

vuntz:
I must admit I'm not sure I understand the rationale for moving this inside the recipe. What was the issue with having it in the template?

rsalevsky:
Since #420 we give anyway a list of packages to the template. My basic idea of this PR is to get rid of the is_ses attribute because the code complexity is too height. And adding the additional check from #422 makes the situation not better. 
While this I moved also the other packages because having the packages at one place makes it easier to maintain the code. Otherwise you have to look at two points for knowing which packages will be installed.

chef/cookbooks/provisioner/recipes/update_nodes.rb
```diff
@@ -225,14 +225,32 @@
...
+              else
+                packages.push("suse-openstack-cloud-release")
+              end
+            end
+
+            if node[:provisioner][:suse][:storage_available]
```
vuntz: So I don't think the logic is correct. If Cloud and Storage are available, we don't want to install ses-release automatically.

vuntz:
@rsalevsky to be honest, I'm not sold on the idea that this should be moved from the template to the recipe. And I'm still a bit confused as to what is wrong with the original code :-)

rsalevsky:
@vuntz The code itself is fine but I think having the if else stuff in the template is not correct. In my opinion a template should not contain to much code especially not when this can be done easily in the script itself. It makes the template unclean and it's hard to see whats going on.

So it's just my personal opinion so feel free to close the PR when I'm wrong.

vuntz:
> So it's just my personal opinion so feel free to close the PR when I'm wrong.

I don't think you're "wrong", I think it's a matter of taste. So let's see what others think before deciding which way to go.

chef/cookbooks/provisioner/recipes/update_nodes.rb
tboerger: This should be only a check for is_ses

tboerger:
I like to keep templates simple, but maybe it's already enough to put the packages with a loop into the template.

PR #431, opened 2015-06-24
Opening port 445 for samba commands

ovisan:
Modified the windows unattended file to open port 445 for samba commands
https://trello.com/c/ZnrdM70i/382-3-p10-make-reinstall-work-for-windows-nodes

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/unattended.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/unattended.xml.erb
@@ -151,6 +151,10 @@
                     <CommandLine>shutdown /r /t 0</CommandLine>
                     <Order>17</Order>
                 </SynchronousCommand>
+                <SynchronousCommand wcm:action="add">
+                    <CommandLine>netsh advfirewall firewall set rule name="File and Printer Sharing (SMB-In)" new enable=yes</CommandLine>
+                    <Order>18</Order>
+                </SynchronousCommand>
             </FirstLogonCommands>
             <UserAccounts>
                 <AdministratorPassword>


chef/cookbooks/provisioner/templates/default/unattended.xml.erb
```diff
@@ -88,79 +88,83 @@
...
                 </SynchronousCommand>
+                <SynchronousCommand wcm:action="add">
+                    <CommandLine>shutdown /r /t 0</CommandLine>
+                    <Order>18</Order>
+                </SynchronousCommand>
             </FirstLogonCommands>
             <UserAccounts>
                 <AdministratorPassword>
-                    <Value><%=@admin_password%></Value>
+                    <Value><%=@admin_pass%></Value>
```
tboerger: This should stay as ```admin_password```

chef/cookbooks/provisioner/templates/default/unattended.xml.erb
```diff
@@ -88,79 +88,83 @@
...
                     <PlainText>true</PlainText>
                 </AdministratorPassword>
             </UserAccounts>
             <AutoLogon>
                 <Password>
-                    <Value><%=@admin_password%></Value>
+                    <Value><%=@admin_pass%></Value>
```
tboerger: This should stay as ```admin_password```

tboerger:
A small wrong change but beside that LGTM

ovisan:
i didn't touch that part, just the second synccommand

tboerger:
Then you took a wrong base for your feature branch :)

ovisan:
Think is better now :-)

tboerger:
:+1:

chef/cookbooks/provisioner/templates/default/unattended.xml.erb
```diff
@@ -88,68 +88,72 @@
                     <Order>1</Order>
                 </SynchronousCommand>
                 <SynchronousCommand wcm:action="add">
+                    <CommandLine>netsh advfirewall firewall set rule name="File and Printer Sharing (SMB-In)" new enable=yes</CommandLine>
+		    <Order>2</Order>
```
vuntz: Why not add it at the end (to avoid changing the order of everything else)? Does it matter? (although I won't object to the patch just for this, just curious)

vuntz:
I'm kind of +1, but I could imagine that some customers wouldn't want this. So I'm unsure if we should always enable this, or add an option for people who want some to avoid opening ports that are not absolutely required.

Any other opinion?

chef/cookbooks/provisioner/templates/default/unattended.xml.erb
ovisan: It doesn't matter, but i grouped the netsh commands together. I could have added this as last in the Order.

tboerger:
> Any other opinion?

But it is required to get the Reboot buttons in the crowbar UI working :)

tboerger:
@ovisan some comment within the commit message with an explanation for the change should be pretty good.

ovisan:
Nope. To reboot from linux it's either samba commands or WMI

ovisan:
@tboerger the explanation is in the title, it's needed for the samba commands. I can expland on that if needed.

vuntz:
> > Any other opinion?
> 
> But it is required to get the Reboot buttons in the crowbar UI working :)

Sure, but this feature is hardly essential, while some people might prefer to be more secure. So the question is whether we should be proactive in addressing their need or waiting for the request to come in.

tboerger:
> @tboerger the explanation is in the title, it's needed for the samba commands. I can expland on that if needed.

In the commit message please! Nobody will remember this pull request.

ovisan:
@vuntz  moved the command to Order 18 from Order 2

ovisan:
@tboerger expanded more info in the commit message

tboerger:
:+1:

tboerger:
@vuntz For now I would take it like it is because it covers the acceptance criteria of the trello card. Maybe you want to create a followup card to extend this functionality. To make it more secure I already suggested an idea.

* Add an option below windows admin password for open/close the port
* If port is closed, set an attribute with chef to execute the reboot or power cycle directly on the node
* If port is open, try to execute the command through samba with the admin password
* If command fails (because password have been changed on the node) fall back to the chef attribute like it is done if the port is closed.

That should be the most flexible solution.

vuntz:
> @vuntz For now I would take it like it is because it covers the acceptance criteria of the trello card. Maybe you want to create a followup card to extend this functionality.

Fair enough.

> To make it more secure I already suggested an idea. [...]

Yeah, using an attribute for reboot/shutdown is actually the previous idea for implementing this feature. I think it makes sense.

vuntz:
+1

PR #434, opened 2015-06-29
add some support for SLES-11-SP4

bmwiedemann:
This is mostly copying what we did for SP3

code is slightly tested (I installed my NUC with it by mounting the SP4-GM DVD onto the suse-11.3 install dir).

### DIFF
--- a/chef/cookbooks/provisioner/libraries/repositories.rb
+++ b/chef/cookbooks/provisioner/libraries/repositories.rb
@@ -21,7 +21,7 @@ class Provisioner
         case type
         when :cloud
           case version
-          when "11.3"
+          when /^11\.[34]$/
             %w(
               Cloud
               Cloud-PTF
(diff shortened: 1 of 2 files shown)


tboerger:
:+1:

rsalevsky:
+1

vuntz:
Code-wise, +1. But before merging: I'm wondering if we should also have a PR for install-suse-cloud at the same time (since we usually do the two changes together).

tboerger:
> Code-wise, +1. But before merging: I'm wondering if we should also have a PR for install-suse-cloud at the same time (since we usually do the two changes together).

It makes sense to integrate it into install-chef-suse, but anyway it doesnt block us from merging this.

bmwiedemann:
I also think that the install-chef-suse change is independent of this one. Maybe it would require this change but not the other way round.

PR #440, opened 2015-07-13
export EDITOR variable

bmwiedemann:
without this, the variable is only available in bash
but not in sub-processes such as
knife node edit $NODE
or
crowbar nova proposal edit default


dirkmueller:
+1

jsuchome:
+1

PR #446, opened 2015-08-04
Run tftpd as tftp user

vuntz:
This is the same as what is done when using the tftpd service file. This
also ensures that tftpd doesn't use the nobody user, who might not have
access to some resources.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/tftp.erb
+++ b/chef/cookbooks/provisioner/templates/default/tftp.erb
@@ -11,5 +11,5 @@ service tftp
        flags                   = IPv6 IPv4
        user                    = root
        server                  = /usr/sbin/in.tftpd
-       server_args             = -m /etc/tftpd.conf -s <%=@tftproot%>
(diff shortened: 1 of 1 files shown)


dirkmueller:
that doesn't fix a thing since apache2 provisioner also runs as wwwrun:www user and can not read the /srv/tftpboot directory anymore

vuntz:
@dirkmueller right, I mentioned that in the card. I guess we need to make /srv/tftpboot o+rx in the cookbook (for now -- ideally, we would move what apache uses out of there, to some place like /srv/www/provisioner).

MaximilianMeister:
see https://github.com/crowbar/barclamp-provisioner/pull/447

tboerger:
Codewise :+1:, but I don't know what the best option should be.

dirkmueller:
filed http://bugzilla.suse.com/show_bug.cgi?id=940608 for the tftp messup, this change is no longer necessary.

PR #447, opened 2015-08-05
Work around tftp packaging change on SLE12 that breaks Provisioner

dirkmueller:
A questionable change in the SLE12 tftp package enforces
a permissions on /srv/tfpboot that breaks crowbar and the apache
provisioner. This has been filed as
https://bugzilla.suse.com/show_bug.cgi?id=940608

### DIFF
--- a/chef/cookbooks/provisioner/recipes/setup_base_images.rb
+++ b/chef/cookbooks/provisioner/recipes/setup_base_images.rb
@@ -240,6 +240,15 @@ when "redhat","centos"
   package "tftp-server"
 when "suse"
   package "tftp"
+
+  # work around change in bnc#813226 which breaks
+  # read permissions for nobody and wwwrun user
+  directory tftproot do
+    recursive true
+    mode 0755
+    owner "root"
+    group "root"
+  end
 end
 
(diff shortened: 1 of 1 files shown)


chef/cookbooks/provisioner/recipes/setup_base_images.rb
```diff
@@ -240,6 +240,15 @@
   package "tftp-server"
 when "suse"
   package "tftp"
+
+  # work around change in bnc#813226 which breaks
+  # read permissions for nobody and wwwrun user
+  directory tftproot do
+    recursive true
```
MaximilianMeister: does the recursive really make sense here? wouldn't this enforce ownership, group and permissions on all files/folders underneath?

dirkmueller: https://docs.chef.io/resource_directory.html

its recursively creating the directory (aka first /srv, then /srv/tfpboot). Also the permissions are only set on the leaf directory, the rest stays untouched.

tboerger:
Codewise :+1:

MaximilianMeister:
+1

PR #449, opened 2015-08-13
Fix power off state for reboot/shutdown

toabctl:
When a node is rebooted while the systemd service crowbar_join.service
is still in state "activating", the job is just cancelled without executing
the "ExecStop" command. Creating a new service which is responsible to set
the power state via "crowbar_join --stop" during reboot/shutdown solves this
problem.

### DIFF
--- a/chef/cookbooks/provisioner/files/default/crowbar_join.service
+++ b/chef/cookbooks/provisioner/files/default/crowbar_join.service
@@ -1,13 +1,13 @@
 [Unit]
 Description=Synchronize with Crowbar administration server
-After=network.target syslog.target remote-fs.target sshd.service neutron-ovs-cleanup.service xend.service libvirtd.service
+After=crowbar_notify_shutdown.service
 Before=chef-client.service
 
 [Service]
 Type=oneshot
 RemainAfterExit=true
 ExecStart=/usr/sbin/crowbar_join --start
-ExecStop=/usr/sbin/crowbar_join --stop
+ExecStartPost=/usr/bin/echo "crowbar_join --start done"
 
 [Install]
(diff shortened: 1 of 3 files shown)


toabctl:
btw. the problem is easy reproducable with the current code. 
First possible reproducer:
1) run `systemctl start crowbar_join`
2) run `systemctl stop crowbar_join` from a second console. This shows that the `ExecPost` is not executed

Second reproducer:
1) Reboot a node and wait until the node is in `Readying` state
2) Reboot the node again. State will stay in `Readying`

tboerger:
I don't like the naming ```crowbar_join_stop```, but codewise LGTM. Damn systemd...

dirkmueller:
"crowbar_leave.service" perhaps instead of crowbar_join_stop?

dirkmueller:
LGTM

toabctl:
> crowbar_leave.service perhaps instead of crowbar_join_stop?

But we are not leaving crowbar (the machine is still registered in crowbar). But I agree that the naming is not nice.

dirkmueller:
+1

dirkmueller:
> But we are not leaving crowbar (the machine is still registered in crowbar). But I agree that the naming
> is not nice.

crowbar_join also does more than just joining crowbar. another try: ```crowbar_notify_shutdown.service```

toabctl:
> crowbar_join also does more than just joining crowbar. another try: crowbar_notify_shutdown.service

`crowbar_notify_shutdown.service` sounds good to me. Done in https://github.com/crowbar/barclamp-provisioner/commit/9351f63e10370b4da06e912f09a95a9cc129d123

PR #453, opened 2015-09-02
wipe out raid metadata

jsuchome:
based on https://github.com/crowbar/crowbar-core/pull/15


dirkmueller:
+1 (would be nice to cherry-pick the exact commit from master, but I'm fine with it as it is)

rsalevsky:
+1

PR #454, opened 2015-09-03
Create '/boot' partition for installation (bsc#935283)

dguitarbite:
SLE11 nodes have issues while booting/installation for harddisk size > 2
TB. Create a boot partition of type gpt with embeded mbr to allow using
bigger disks for non-raid type.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -27,6 +27,15 @@
     </signature-handling>
     <storage/>
   </general>
+  <% if @target_platform_version.to_f < 12 -%>
+    <bootloader>
+      <global>
+        <activate>true</activate>
+        <boot_boot>true</boot_boot>
+        <boot_mbr>true</boot_mbr>
+      </global>
+    </bootloader>
+  <% end -%>
   <add-on>
     <add_on_products config:type="list">
     <% @repos.keys.sort.each do |name| %>
@@ -96,7 +105,52 @@
           <device>/dev/<%= @boot_device %></device>
         <% end %>
         <initialize config:type="boolean">true</initialize>
-        <partitions config:type="list"/>
+        <% if @target_platform_version.to_f < 12 -%>
+          <partitions config:type="list">
+            <partition>
+              <create config:type="boolean">true</create>
+              <filesystem config:type="symbol">ext3</filesystem>
+              <format config:type="boolean">true</format>
+              <mount>/boot</mount>
+              <size>auto</size>
+            </partition>
+            <partition>
+              <create config:type="boolean">true</create>
+              <filesystem config:type="symbol">swap</filesystem>
+              <format config:type="boolean">true</format>
+              <mount>swap</mount>
+              <size>auto</size>
+            </partition>
+            <partition>
+              <create config:type="boolean">true</create>
+              <format config:type="boolean">true</format>
+              <mount>/</mount>
+              <size>auto</size>
+            </partition>
+          </partitions>
+        <% else -%>
+          <disklabel>gpt</disklabel>
+          <partitions config:type="list">
+            <partition>
+              <create config:type="boolean">true</create>
+              <partition_id config:type="integer">263</partition_id>
+              <size>10M</size>
+            </partition>
+            <partition>
+              <create config:type="boolean">true</create>
+              <filesystem config:type="symbol">swap</filesystem>
+              <format config:type="boolean">true</format>
+              <mount>swap</mount>
+              <size>auto</size>
+            </partition>
+            <partition>
+              <create config:type="boolean">true</create>
+              <format config:type="boolean">true</format>
+              <mount>/</mount>
+              <size>auto</size>
+            </partition>
+          </partitions>
+        <% end -%>
         <type config:type="symbol">CT_DISK</type>
         <use>all</use>
       </drive>


dguitarbite:
Sending this PR to tex because it affects SLE11 nodes only. For SLE12 nodes, we do not need gpt with embedded mbr. No need to add it for Cloud6 as its SLE12 only (grub version 2.0).

bmwiedemann:
:+1: change looks good and was probably tested

rsalevsky:
@dguitarbite Can you split the echo change in a separate commit?

vuntz:
> Sending this PR to tex because it affects SLE11 nodes only. For SLE12 nodes, we do not need gpt with embedded mbr.

Then it means that the new bits that are added here should be for autoyast for SLE11 nodes only; as we do support SLE12 nodes in tex, there needs to be some logic for this.

dguitarbite:
@vuntz You are right. I am trying to get the right patch for SLE11/12 nodes. If I get my generic XML working then we do not need chef logic to switch between nodes.

WIP: Creating a fix for SLE12 nodes.

dguitarbite:
This PR has a /boot partition for considering SLE11 nodes. Meanwhile it should also work for SLE12 nodes. For cloud 6 we dont need this and should be fine with using `bios_boot` partition.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -96,8 +104,38 @@
           <device>/dev/<%= @boot_device %></device>
         <% end %>
         <initialize config:type="boolean">true</initialize>
-        <partitions config:type="list"/>
         <type config:type="symbol">CT_DISK</type>
+        <disklabel>gpt</disklabel>
```
dirkmueller: do we have to enforce the disklabel to GPT? I'm afraid that this might break non-GPT capable setups (not sure if we still care in 2015, but since this is a maintenance branch..)

dguitarbite: For larger disks, we need to specify gpt disklabel. I am not so sure if we break the non-GPT setups ( AFAIK, it should work with GPT. Not sure why this breaks non-GPT setups?).

dguitarbite:
@dirkmueller I have confirmed and tested the same without gpt partition. Its kind of "tricking" the system by keeping the /boot partition in the first 2TB sectors of the disk and after booting, the entire disk is usable. So we do not need to have extensive checks for larger and smaller disks and should work for older boot-loader firmwares.

dirkmueller:
:+1:

rsalevsky:
+1

dguitarbite:
Is this patch ready to be merged?

dirkmueller:
What I#m wondering though is why this code is so much different than the code we have in master now?

thinking about it, maybe we just want to backport the sle12-only solution under a big "if sle12" if and leave the sle11 part unfixed?

dguitarbite:
@dirkmueller The code is different because we need to consider Grub1 for SLE11 which makes it difficult to simply backport the patch from master.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -27,6 +27,13 @@
     </signature-handling>
     <storage/>
   </general>
+  <bootloader>
```
dirkmueller: should be inside the if section?

dguitarbite: @dirkmueller This is the same for SLE11/12. IMO if section would be overkill here.

tboerger: Overkill for a simple if?

dguitarbite: Meaning, the XML tags remain the same for SLE11 and SLE12. I do not see any use for if/else statement here.

dguitarbite: I mean to say that this is optional for SLE12, its not explicitly required to have this section.

vuntz:
@dguitarbite just to be 100% sure: it is intended to change the bit related to RAID on SLE12?

dguitarbite:
@vuntz Nope, I think its just tracking the indents. I have not changed any code for raid.

vuntz:
> @vuntz Nope, I think its just tracking the indents. I have not changed any code for raid.

Hrm, you did :-) Just look at this URL: https://github.com/crowbar/barclamp-provisioner/pull/454/files?w=1

dguitarbite:
@vuntz Yes! I have modified the raid setup for SLE11 to accommodate for HDD > 2TB. Sorry for missing this part out.

dguitarbite:
Removed code for RAID. I will address it in another patch.

dguitarbite:
@vuntz @dirkmueller Can we get this merged today if possible?

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -27,6 +27,15 @@
     </signature-handling>
     <storage/>
   </general>
+  <% if @target_platform_version.to_f <= 12 %>
```
dirkmueller: isn't that a complicated way to say "if true" ?

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -96,7 +105,52 @@
           <device>/dev/<%= @boot_device %></device>
         <% end %>
         <initialize config:type="boolean">true</initialize>
-        <partitions config:type="list"/>
+        <% if @target_platform_version.to_f <= 12 %>
```
dirkmueller: isn't that a complicated way to say "if true" ?

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
```diff
@@ -96,7 +105,52 @@
...
+          <partitions config:type="list">
+            <partition>
+              <create config:type="boolean">true</create>
+              <filesystem config:type="symbol">ext3</filesystem>
```
dirkmueller, line 112: do we need to hardcode the filesytem? I would have expected that this is just the empty list like before.

dguitarbite, line 112: For SLE11 yes, for disks > 2TB its more reliable IMO. While testing this manually I faced issues for the same.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
dguitarbite: I see what you mean now, Ill fix this.

chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
dirkmueller: we don't have that in crowbar-core/master so we shouldn't add it to the  SLE12 case on the backport then

dirkmueller:
+1, remaining nitpick is change of "%>" to "-%>" to remove the extra newlines.

jsuchome:
+1 (looks good, I have not tested)

PR #455, opened 2015-09-03
Create '/boot' partition for installation (bsc#935283)

dguitarbite:
SLE11 nodes have issues while booting/installation for harddisk size > 2
TB. Create a boot partition of type gpt with embeded mbr to allow using
bigger disks for non-raid type.


rsalevsky:
master is not longer maintained please read the README.md

PR #457, opened 2015-09-08
For large disks, we're gonna need to create BIOS boot partition

jsuchome:
backport of https://github.com/crowbar/crowbar-core/pull/30

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
+++ b/chef/cookbooks/provisioner/templates/default/autoyast.xml.erb
@@ -108,6 +108,11 @@
           <initialize config:type="boolean">true</initialize>
           <partitions config:type="list">
             <partition>
+              <create config:type="boolean">true</create>
+              <partition_id config:type="integer">263</partition_id>
+              <size>10M</size>
+            </partition>
+            <partition>
               <mount>swap</mount>
               <size>auto</size>
             </partition>


dguitarbite:
@jsuchome Just a suggestion to rephrase the commit message "Create bios_boot partition for disks bigger than 4TB size" ... although I do not see any checks for disk size. Rather "Create bios_boot partition to consider bigger RAID disks".

vuntz:
Can't really say if that's correct or not, but nothing obviously wrong. So +0.5 from me because it fixes the issue according to the bug.

dguitarbite:
@vuntz Its probably the best possible replacement for /boot partition since GRUB 2 onwards.

jsuchome:
@dguitarbite I've merged the master version, do you thinnk you can give +1 here as well?

dguitarbite:
+1 LGTM

@jsuchome Although I would suggest a small nitpick in your commit message, can you please add the reference to the other PR and mention it as backport?

rsalevsky:
+1

dguitarbite:
I am not sure if this will work with sle11 as it has grub 1 and bios_boot
partition may not work. Can you verify this once? It should work for sle12
though.

jsuchome:
For SLE12 it works, but you're right, I should test it also for Cloud5 and SLE11 nodes

jsuchome:
@dguitarbite just tested with SLE11 nodes of Clood5: works fine.

dguitarbite:
Awesome, in that case I'll update my patch too. In this case I believe we
do not need a /boot partition.

PR #459, opened 2015-10-15
Do not write IP address of the node when it is not known yet.

jsuchome:
partially backporting crowbar/crowbar-core#78

Depends on https://github.com/crowbar/barclamp-deployer/pull/253


tboerger:
LGTM codewise

vuntz:
+1

PR #466, opened 2016-05-19
[tex] bsc#980569 sntp changes

MaximilianMeister:
https://bugzilla.suse.com/show_bug.cgi?id=980569
It seems that there has been some changes in a recently released sntp

the `-P` option has been dropped, and `-r` is now `-S`

for further reading refer to the commit messages

this is currently an issue in the cloud 5 ci jobs:

sntp

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -133,7 +133,7 @@ sync_time() {
     local tries_left=120
 
 <% if @target_platform_version.to_f < 12.0 %>
-    SNTP_OPTS="-P no -r"
+    SNTP_OPTS="-r"
 <% else %>
     SNTP_OPTS="-s"
 <% end %>
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -133,7 +133,7 @@ sync_time() {
     local tries_left=120
 
 <% if @target_platform_version.to_f < 12.0 %>
-    SNTP_OPTS="-r"
+    SNTP_OPTS="-S"
 <% else %>
     SNTP_OPTS="-s"
 <% end %>
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -380,7 +380,7 @@ export DOMAIN
 export HOSTNAME
 
 <% if @target_platform_version.to_f < 12.0 -%>
-ntp="sntp -r $NTP_SERVERS"
+ntp="sntp -S $NTP_SERVERS"
 <% else -%>
 ntp="sntp -s $NTP_SERVERS"
 <% end -%>
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -380,7 +380,7 @@ export DOMAIN
 export HOSTNAME
 
 <% if @target_platform_version.to_f < 12.0 -%>
-ntp="sntp -P no -r $NTP_SERVERS"
+ntp="sntp -r $NTP_SERVERS"
 <% else -%>
 ntp="sntp -s $NTP_SERVERS"
 <% end -%>
--- a/updates/control.sh
+++ b/updates/control.sh
@@ -99,7 +99,7 @@ export DHCP_STATE MYINDEX ADMIN_ADDRESS BMC_ADDRESS BMC_NETMASK BMC_ROUTER ADMIN
 export ALLOCATED HOSTNAME CROWBAR_KEY CROWBAR_STATE
 
 if is_suse; then
-    ntp="sntp -P no -r $ADMIN_IP"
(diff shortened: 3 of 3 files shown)


MaximilianMeister:
just updated the commit messages to include the bug number
old gating was https://ci.suse.de/job/openstack-mkcloud/23178/console

MaximilianMeister:
jfyi: the nodes seem to be up and running in https://ci.suse.de/job/openstack-mkcloud/23178/console

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -133,7 +133,7 @@ sync_time() {
     local tries_left=120
 
 <% if @target_platform_version.to_f < 12.0 %>
-    SNTP_OPTS="-P no -r"
+    SNTP_OPTS="-S"
```
rhafer, line 136: Is there any reason for using `-S` on SLE11 and `-s` on SLE12 (and newer). AFAIK both release are using the same ntp packages now.

rhafer:
-1 
We cannot expect that all customers will have the updated ntp version (and a sleshammer build with the new ntp) installed. So I guess we need to somehow implement this in a backwards compatible way. And the "old" variant of sntp does not no anything about "-s/-S" :(

MaximilianMeister:
> We cannot expect that all customers will have the updated ntp version (and a sleshammer build with the new ntp) installed. So I guess we need to somehow implement this in a backwards compatible way. And the "old" variant of sntp does not no anything about "-s/-S" :(

good points there. but not sure how to achieve that, so could we maybe include a check for the package version here?

nicolasbock:
Why are we using `sntp` as opposed to say `ntpdate` or `ntpd` here? Would one of the other provide a more consistent command line argument situation?

MaximilianMeister:
> Why are we using sntp as opposed to say ntpdate or ntpd here? Would one of the other provide a more consistent command line argument situation?

could work too, but this needs to be investigated. Unfortunately I cannot follow up on this because of vacation, but we decided to create a trello card for that. feel free to close thise once it is not needed anymore

rhafer:
> Why are we using sntp as opposed to say ntpdate or ntpd here? Would one of the other provide a more consistent command line argument situation?

The reason is that ntpdate was not available on SLES11 (IIRC it was just a shell script telling the user to use  sntp). Not sure if it's possible to do a one-time sync with ntpd.

nicolasbock:
Yes, you can use `ntpd -gq`, which behave just like calling `ntpdate`.

rhafer:
> good points there. but not sure how to achieve that, so could we maybe include a check for the package version here?

Checking the package version does not work, as the sledgehammer image does not have the rpm database nor the tooling for that. If `ntpd -gq <ntpservers>` really behaves like ntpdate/sntp everywhere I guess it would be ok to switch to that.

nicolasbock:
> Checking the package version does not work, as the sledgehammer image
> does not have the rpm database nor the tooling for that. If |ntpd -gq
> <ntpservers>| really behaves like ntpdate/sntp everywhere I guess it
> would be ok to switch to that.

How far back in terms of SLES do we support? I could go on a little
expedition and find out whether the ntpds on all supported SLES versions
work with those command line arguments.

rhafer:
The oldest SLES we'd have to expect is SLE-11-SP3 GA (without updates installed) I'd say.

nicolasbock:
On 05/20/2016 12:07 PM, Ralf Haferkamp wrote:

> The oldest SLES we'd have to expect is SLE-11-SP3 GA (without updates
> installed) I'd say.

Ok, thanks. That narrows it down quite a bit. I'll go look.

nicolasbock:
We ship ntpd-4.2.8p6 in SLE-11-SP3 GA. This version supports `gq`. We could therefore replace all calls that do a onetime set of the clock at boot with `ntpd -gq`.

rhafer:
> We ship ntpd-4.2.8p6 in SLE-11-SP3 GA

Actually  SLE-11-SP3 GA has ntp 4.2.4something (4.2.8 was just recently released as an update IIRC, that is exactly what's causing the issues the PR is trying to address)

aplanas:
4.2.4p8

nicolasbock:
> ```
> We ship ntpd-4.2.8p6 in SLE-11-SP3 GA
> ```
> 
> Actually SLE-11-SP3 GA has ntp 4.2.4something (4.2.8 was just recently
> released as an update IIRC, that is exactly what's causing the issues
> the PR is trying to address)

Yes, you are right. I checked and ntp-4.2.4p8 also supports the `-gq`
options.

rhafer:
> I checked and ntp-4.2.4p8 also supports the `-gq`options

But it seems you can NOT specify a server on the command line in the same way as for the 4.2.8 version :(

rhafer:
Since this is blocking CI for Tex and @nicolasbock and @MaximilianMeister  are not there today (as discussed in todays standup) here's a reworked (more generic) PR: https://github.com/crowbar/barclamp-provisioner/pull/467

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
bmwiedemann, line 136: for some reason, I only see ntp-4.2.4 on SLE11-SP3 but ntp-4.2.8 on SLE12-GA
- is in LTSS and cloud5 repo

PR #467, opened 2016-05-23
Use ntpdate on systems with newer ntpd (bsc#980569)

rhafer:
With a recent update to ntpd-4.2.8p6 the commandline options of sntp changed in
a backwards incompatible way. Try to find out with version of ntp we're running
and uses sntp for older releases (4.2.4 didn't contain a working ntpdate) and
ntpdate for newer releases to do the initial time sync.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=980569

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
@@ -132,14 +132,19 @@ sync_time() {
 
     local tries_left=120
 
-<% if @target_platform_version.to_f < 12.0 %>
-    SNTP_OPTS="-P no -r"
-<% else %>
-    SNTP_OPTS="-s"
-<% end %>
+    # ntpd 4.2.4 (e.g. on SLES11-SP3 GA) didn't contain a working ntpdate. Use sntp
+    # there. ntpdate got reintroduced with a newer ntpd versions (4.2.8x) where the
+    # sntp commandline interface changed in an incompatible way. Use ntpdate on
+    # such systems (i.e. SLES-11-SP3 with updates and SLES12)
+    if ntpd --version | grep -q "4\.2\.4"; then
+        ntp="sntp -P no -r"
+    else
+        ntp="/usr/sbin/ntpdate -u"
+    fi
+
     while [[ $tries_left > 0 ]] ; do
         for ntpserver in $VALID_NTP_SERVERS ; do
-            if sntp $SNTP_OPTS $ntpserver; then
+            if $ntp $ntpserver; then
                 break 2
             fi
         done
--- a/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
+++ b/chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
@@ -379,11 +379,16 @@ hostname "$HOSTNAME"
 export DOMAIN
 export HOSTNAME
 
-<% if @target_platform_version.to_f < 12.0 -%>
-ntp="sntp -P no -r $NTP_SERVERS"
-<% else -%>
-ntp="sntp -s $NTP_SERVERS"
-<% end -%>
+# ntpd 4.2.4 (e.g. on SLES11-SP3 GA) didn't contain a working ntpdate. Use sntp
+# there. ntpdate got reintroduced with a newer ntpd versions (4.2.8x) where the
+# sntp commandline interface changed in an incompatible way. Use ntpdate on
+# such systems (i.e. SLES-11-SP3 with updates and SLES12)
+if ntpd --version | grep -q "4\.2\.4"; then
+    ntp="sntp -P no -r $NTP_SERVERS"
+else
+    ntp="/usr/sbin/ntpdate -u $NTP_SERVERS"
+fi
+
 # Make sure date is up-to-date
 until $ntp; do
   echo "Waiting for NTP server(s)"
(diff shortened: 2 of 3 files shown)


rhafer:
Note, I tested this manually against GM5+up as well to verify that it really works with the currently released sleshammer.

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
```diff
@@ -132,14 +132,19 @@ sync_time() {
...
-    SNTP_OPTS="-s"
-<% end %>
+    # ntpd 4.2.4 (e.g. on SLES11-SP3 GA) didn't contain a working ntpdate. Use sntp
+    # there. ntpdate got reintroduced with a newer ntpd versions (4.2.8x) where the
+    # sntp commandline interface changed in an incompatible way. Use ntpdate on
+    # such systems (i.e. SLES-11-SP3 with updates and SLES12)
+    if ntpd --version | grep -q "4\.2\.4"; then
+        ntp="sntp -P no -r"
+    else
+        ntp="/usr/sbin/ntpdate"
```
vuntz: We use `/usr/sbin/ntpdate -u` in Cloud 6 (to avoid the `the NTP socket is in use, exiting` issue),  should we do the same here?

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -379,11 +379,16 @@ hostname "$HOSTNAME"
...
-ntp="sntp -s $NTP_SERVERS"
-<% end -%>
+# ntpd 4.2.4 (e.g. on SLES11-SP3 GA) didn't contain a working ntpdate. Use sntp
+# there. ntpdate got reintroduced with a newer ntpd versions (4.2.8x) where the
+# sntp commandline interface changed in an incompatible way. Use ntpdate on
+# such systems (i.e. SLES-11-SP3 with updates and SLES12)
+if ntpd --version | grep -q "4\.2\.4"; then
+    ntp="sntp -P no -r $ADMIN_IP"
+else
+    ntp="/usr/sbin/ntpdate $ADMIN_IP"
```
vuntz: Same comment.

chef/cookbooks/provisioner/templates/default/crowbar_join.suse.sh.erb
vuntz: Btw, in Cloud 6, we pass all ntp servers at once to ntpdate, instead of doing a loop for each of them.

rhafer: > We use /usr/sbin/ntpdate -u in Cloud 6 (to avoid the the NTP socket is in use, exiting issue), should we do the same here?

Yeah, I'll add that.

> Btw, in Cloud 6, we pass all ntp servers at once to ntpdate, instead of doing a loop for each of them.

AFAICS the old sntp does not allow passing more than one server at a time :/

vuntz: > AFAICS the old sntp does not allow passing more than one server at a time :/

Right, but you can do it just for ntpdate, I guess. Not sure it matters much, though.

rhafer: > Right, but you can do it just for ntpdate, I guess. Not sure it matters much, though.

I am a bit reluctant to do that for tex honestly. I think it only complicates the patch for very little benefit.

rhafer:
@vuntz updated, but didn't change the multiple server case (see my comments for that)

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
```diff
@@ -379,11 +379,16 @@ hostname "$HOSTNAME"
...
-ntp="sntp -s $NTP_SERVERS"
-<% end -%>
+# ntpd 4.2.4 (e.g. on SLES11-SP3 GA) didn't contain a working ntpdate. Use sntp
+# there. ntpdate got reintroduced with a newer ntpd versions (4.2.8x) where the
+# sntp commandline interface changed in an incompatible way. Use ntpdate on
+# such systems (i.e. SLES-11-SP3 with updates and SLES12)
+if ntpd --version | grep -q "4\.2\.4"; then
+    ntp="sntp -P no -r $ADMIN_IP"
+else
+    ntp="/usr/sbin/ntpdate -u $ADMIN_IP"
```
vuntz: This should really work on `$NTP_SERVERS`; that being said, the sledgehammer script is also wrong, so not sure it matters much...

vuntz:
One small comment; it might not be important, so you get my +1.

chef/cookbooks/provisioner/templates/suse/crowbar_register.erb
rhafer: Oh. True :(

rhafer: Adjusted this. It's still not 100% correct, but at least not more broken than before :/

rhafer:
@rsalevsky Want to give this a review? This should unbreak the D:C:5 gating.

rsalevsky:
+1 In case there is still something broken we can create follow up PR.

nicolasbock:
:+1:

